PR #1, opened 2015-07-23
Update routing

dustinlarimer:
This PR moves the explorer module's routing to whatever is provided within the `appBase` config option. I'm not 100% sure if this is the appropriate way to fix this issue, but it works now (for the demo).


hex337:
This seems good, will let Joe do the official +1.

josephwegner:
+1 thanks dustin

dustinlarimer:
Closing, as this isn't the proper place to make this change.

PR #16, opened 2015-08-14
merge public and private repos

josephwegner:
@hex337 @dustinlarimer @aroc 

This brings over all the recent changes to the private fork of Explorer.

The only things that look less than open-source-happy here is that we have some terrarium/keen-web specific gulp tasks. I'm fine with letting those show up over here, until we work out a solution that will make sense for splitting them.


hex337:
Blah, I started a task that was splitting up the gulp tasks into their own files and never finished it, I should do that.

josephwegner:
If you think it should block, we can dive into that... but doesn't seem worth blocking for now.

hex337:
Nope, don't block this!

On Fri, Aug 14, 2015 at 12:52 PM, Joe Wegner <notifications@github.com>
wrote:

> If you think it should block, we can dive into that... but doesn't seem
> worth blocking for now.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/keen/explorer/pull/16#issuecomment-131219092>.
>



-- 
-Alex
Keen IO Open Source devAdvocate

aroc:
Looks good! :+1:

PR #18, opened 2015-08-15
Visit Duration now defaults to pie chart

sfalkoff:
Sets the default query view for `Visit Duration` to a pie chart.

screen shot 2015-08-14 at 5 03 37 pm


spraetz:
:+1:

PR #21, opened 2015-08-17
add an option to set a max_age for all queries

josephwegner:
@aroc @hex337 @dustinlarimer 

This adds a configuration property to put a `max_age` on all queries.

My only worry here is that it _seems_ like there is some structure in `app.js` for a `config.options` object, which should hold this kind of thing... However, those options gets passed to the Component. I couldn't find a way to go through the component, but still access the `max_age` inside of `ExplorerUtils.queryJSON`.

If y'all can think of a cleaner way to do this, please let me know.

### DIFF
--- a/client/js/app/app.js
+++ b/client/js/app/app.js
@@ -36,7 +36,7 @@ function App(config) {
 
   // Create the main active explorer
   var explorerAttrs = _.assign(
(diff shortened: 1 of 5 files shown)


josephwegner:
bump

aroc:
Closing as we're moving away from `max_age` in favor of saved/cached queries.

PR #22, opened 2015-08-17
add favicon to landing page

elof:


josephwegner:
@elof this repo has its own favicon in it, but it seems like it was just never hooked up..

Could you point to http://keen.github.io/explorer/favicon.ico, instead of using the asset from keen.io?

Thanks!

elof:
Sounds good. I updated the link.

josephwegner:
+1, merging

Issue #25, opened 2015-08-22
Relative Reference of rest.js in client > js > app > modules > persistence > persistence.js


shabeermothi:
Was setting up the keen data explorer using the instructions in readme and encountered an issue while serving the project with gulp.. Below is the trace 

`[00:38:08] Using gulpfile /Users/shabeermothi/projects/keen-data-explorer/explorer/gulpfile.js
[00:38:08] Starting 'build-scripts'...
[00:38:08] Starting 'images'...
[00:38:08] Starting 'fonts'...
[00:38:08] Starting 'build-styles'...
[00:38:08] Starting 'connect'...
[00:38:08] Server started http://localhost:8082
[00:38:08] Finished 'connect' after 238 ms
[00:38:08] Starting 'watch'...
[00:38:08] Finished 'watch' after 135 ms
Autoprefixer's process() method is deprecated and will removed in next major release. Use postcss([autoprefixer]).process() instead
[00:38:08] Finished 'build-styles' after 820 ms
[00:38:08] Finished 'images' after 834 ms
[00:38:10] Finished 'fonts' after 2.2 s
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: module "./rest.js" not found from "/Users/shabeermothi/projects/keen-data-explorer/explorer/client/js/app/modules/persistence/persistence.js"
    at notFound (/Users/shabeermothi/projects/keen-data-explorer/explorer/node_modules/browserify/index.js:812:15)
    at /Users/shabeermothi/projects/keen-data-explorer/explorer/node_modules/browserify/index.js:762:23
    at /Users/shabeermothi/projects/keen-data-explorer/explorer/node_modules/browserify/node_modules/browser-resolve/index.js:250:24
    at /Users/shabeermothi/projects/keen-data-explorer/explorer/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:55:18
    at load (/Users/shabeermothi/projects/keen-data-explorer/explorer/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/Users/shabeermothi/projects/keen-data-explorer/explorer/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /Users/shabeermothi/projects/keen-data-explorer/explorer/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
[...]

PR #26, opened 2015-08-27
fix dependency #25

clayallsopp:
# What's this PR do?

Fixes the repo to work according to the checkout instructions (see #25)

# Where should the reviewer start?

The fix is pretty simple; I also included the changes to the dist, not sure if that's what you want though

# How should this be manually tested?

Follow the instructions in the README - before the PR, it fails, after the PR, it works

# Screenshots (if appropriate)


josephwegner:
Love it! hanks @clayallsopp!

PR #30, opened 2015-10-20
Sync open and private

josephwegner:
@aroc @joannecheng @dustinlarimer 

So, this should sync up what we have done privately... and then we can kill dat private bizness??????!??!?!??!??!?!?!?


josephwegner:
@aroc IMO we should merge this, and do the new changes in a new PR... That'll make the rebase a bit less insanely painful.

Can we do this today, and kill private explorer?

aroc:
Let's do it!

PR #31, opened 2015-10-27
Final oss sync

josephwegner:
@eric OK... I think this is everything now. Tests pass, and it looks pretty solid in my local environment... Shall we do this thang?


josephwegner:
FYI, I pushed the same branch to the private version, under https://github.com/keen/explorer/pull/31

That diff is a bit nicer to read than this one.

PR #37, opened 2015-11-04
Funnel data model

josephwegner:
@aroc 

This adds the `ExplorerStore` logic for switching to & from funnel analysis types. Properties should flow in and out of the steps property when switching analysis types.

There's also a small refactor to use a set of update preparation functions, rather than making the `_update` function super chaotic.

Well, unfortunately, you can't. There is no functional way to actually touch this code, because we don't actually have a funnel analysis_type yet. I'd suggest reading the tests thoroughly. If those make sense, and pass in CI, then we should be golden.

### DIFF
--- a/client/js/app/stores/ExplorerStore.js
+++ b/client/js/app/stores/ExplorerStore.js
@@ -84,6 +84,35 @@ function _getDefaultFilterCoercionType(explorer, filter) {
 }
 
 /**
+ * Runs through the changeset and moves data around if necessary
+ * @param {Object} explorer   The explorer model that is being updated
+ * @param {Object} updates    The updated explorer model
+ * @return {Object}           The new set of updates
+ */
+function _prepareUpdates(explorer, updates) {
+  var newModel = _.assign({}, explorer, updates);
+
+  _removeEmailExtractionFields(explorer, newModel);
+
+  return newModel;
+}
+
+/**
+ * Removes fields from email extractions
+ * @param {Object} explorer   The explorer model that is being updated
+ * @param {Object} newModel   The updated explorer model
+ * @return {Object}           The new set of updates
+ */
+function _removeEmailExtractionFields(explorer, newModel) {
+  if (!ExplorerUtils.isEmailExtraction(newModel)) {
+    newModel.query.latest = null;
+    newModel.query.email = null;
+  }
+
+  return newModel;
+}
+
+/**
  * Looks at updates about to be made to a filter and performs a series of operations to
  * change the filter values depending on the coercion_type, operator and type of property_value.
  * @param  {Object} explorer  The explorer model that owns the filter to be updated
@@ -128,13 +157,8 @@ function _create(attrs) {
 }
 
 function _update(id, updates) {
-  var newModel = _.assign({}, _explorers[id], updates);
-  // If we're no longer doing an email extraction, remove the latest and email field.
-  // FIXME: Does this belong here? Maybe this should be in the onChange callback rather than hard-bound to the model.
-  if (!ExplorerUtils.isEmailExtraction(newModel)) {
-    newModel.query.latest = null;
-    newModel.query.email = null;
-  }
+  var newModel = _prepareUpdates(_explorers[id], updates);
+
   if (updates.id && updates.id !== id) {
     _explorers[updates.id] = newModel;
     delete _explorers[id];
--- a/client/js/app/stores/ExplorerStore.js
+++ b/client/js/app/stores/ExplorerStore.js
@@ -68,6 +68,23 @@ function _defaultGeoFilter() {
   };
 }
 
+function _defaultStep() {
+  return {
+    event_collection: null,
+    actor_property: null,
+    time: {
+      relativity: 'this',
+      amount: 14,
+      sub_timeframe: 'days'
+    },
+    timezone: ProjectUtils.getConstant('DEFAULT_TIMEZONE'),
+    filters: [],
+    optional: false,
+    inverted: false,
+    active: false
+  }
+}
+
 /**
  * Get the default coercion type for this filter based off the filter's property_name's set type
  * in the project schema.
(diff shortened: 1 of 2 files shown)


client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,87 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+    });
+
+    if(typeof(explorer.query.target_property) !== 'undefined' && explorer.query.target_property !== null) {
+      firstStep.actor_property = explorer.query.target_property;
+      delete explorer.query.target_property;
+    }
+
+    newModel.query.steps = [firstStep];
+
+  } else if (newModel.query.analysis_type !== 'funnel' && explorer.query.analysis_type === 'funnel') {
```
josephwegner: I'm thinking maybe this should be a separate function, instead of an `else if`...  thoughts?

I wrote it as one, because the logic is all so similar... but it's a big messy func.

ajwhite: If this is worth anything, as out outside observer trying to familiarize with this file, the separation would be easier on the :eyes: and thought process

client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,87 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+ */
+function _migrateFunnelSteps(explorer, newModel) {
+  var sharedProperties = ['event_collection', 'time', 'timezone', 'filters']
+  if(newModel.query.analysis_type === 'funnel' && explorer.query.analysis_type !== 'funnel') {
+    // Changing TO funnels
+    var firstStep = _defaultStep();
+    firstStep.active = true;
+
+    _.each(sharedProperties, function (key) {
+      if(typeof(explorer.query[key]) !== 'undefined' && explorer.query[key] !== null) {
```
joannecheng: if we're using underscore in this file, we can probably use `_.isUndefined` and `_.isNull` here.

josephwegner: Heh, that's my thought too, but I think @aroc said he prefers to not get too tied to a lib. 

@aroc thoughts?

ajwhite: I'd say that's a good rule if you are trying to replace everything, even simple things like the following with lodash

```js
var foo = foobar.foo;
// vs
var foo = _.get(foobar, 'foo');
```

But since you are using lodash in your project, and you're doing what feels a bit more "manual" than `_.isUndefined` or `_.isNull`, I don't see any reason to not use a simpler syntax that lodash provides.

Unless you plan on leaving lodash anytime soon.

client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,99 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+}
+
+/**
+ * If the query got changed to a funnel, move the step-specific parameters to a steps object.
+ * @param {Object} explorer The explorer model that is being updated
+ * @param {Object} newModel The updated explorer model
+ * @return {Object}         The new set of updates
+ */
+function _migrateToFunnel(explorer, newModel) {
+  var sharedProperties = ['event_collection', 'time', 'timezone', 'filters']
```
aroc: semicolon missing.

client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,99 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+  if(newModel.query.analysis_type === 'funnel' && explorer.query.analysis_type !== 'funnel') {
+    // Changing TO funnels
+    var firstStep = _defaultStep();
+    firstStep.active = true;
+
+    _.each(sharedProperties, function (key) {
+      if(!_.isUndefined(explorer.query[key]) && !_.isNull(explorer.query[key])) {
```
aroc: The query, if it's not a funnel, should *always* have the keys you're looping through here. You could just do:
`if (explorer.queyr[key])`

client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,99 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+ * @param {Object} explorer   The explorer model that is being updated
+ * @param {Object} updates    The updated explorer model
+ * @return {Object}           The new set of updates
+ */
+function _prepareUpdates(explorer, updates) {
+  var newModel = _.assign({}, explorer, updates);
+
+  newModel = _removeEmailExtractionFields(explorer, newModel);
+  newModel = _migrateToFunnel(explorer, newModel);
+  newModel = _migrateFromFunnel(explorer, newModel);
```
aroc: This code reads really weird. The fact that we call `migrateToFunnel` then immediately after `migrateFromFunnel` is a confusing read. It might make sense to do the check in this function and then only call one or the other. While three lines in a row like this is nice and clean, having to read those other functions to understand why these 2 lines don't contradict themselves isn't the best.

These function names are nice and clear to me on their own, so leveraging an if statement here and only calling one or the other would make for a great and clear read of what's going on.

client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,99 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+function _migrateFromFunnel(explorer, newModel) {
+  var sharedProperties = ['event_collection', 'time', 'timezone', 'filters']
+  if (newModel.query.analysis_type !== 'funnel' && explorer.query.analysis_type === 'funnel') {
+    // Changing FROM funnels
+    var activeStep = _.find(explorer.query.steps, function (step) {
+      return step.active
+    });
+
+    _.each(sharedProperties, function (key) {
+      if(!_.isUndefined(activeStep[key])) {
```
aroc: The lodash `isUndefined` function is pretty straightforward:
```
function isUndefined(value) {
      return value === undefined;
    }
```

So for these if statements I think maybe a simple `if(activeStep[key])` or `if(activeStep[key] !== undefined)` should be ok. As we move closer towards hopefully using Babel+ES6 plus, it would be great to drop the need for these utility libs.

client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,99 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+}
+
+/**
+ * If the query got changed from a funnel, move the applicable parameters out to the root query
+ * @param {Object} explorer The explorer model that is being updated
+ * @param {Object} newModel The updated explorer model
+ * @return {Object}         The new set of updates
+ */
+function _migrateFromFunnel(explorer, newModel) {
+  var sharedProperties = ['event_collection', 'time', 'timezone', 'filters']
```
aroc: Maybe move these up into a constant as to avoid having to redeclare inside each funnel migration function?

client/js/app/stores/ExplorerStore.js
```diff
@@ -84,6 +101,99 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
+  return newModel;
+}
+
+/**
+ * Removes fields from email extractions
+ * @param {Object} explorer   The explorer model that is being updated
+ * @param {Object} newModel   The updated explorer model
+ * @return {Object}           The new set of updates
+ */
+function _removeEmailExtractionFields(explorer, newModel) {
```
aroc: Thanks for cleaning my junky code up a bit! :)

aroc:
Ok, check out the notes I left inline.

client/js/app/stores/ExplorerStore.js
josephwegner: We should chat about this as a team. I originally had all of these without lodash, and got feedback to use lodash (which I originally didn't do, because I knew you preferred no lodash).

FWIW, I was doing `typeof(value) !== 'undefined'`, which is harder to read, so maybe that's the root of my issues.

josephwegner: the original convo: https://github.com/keen/explorer/pull/37#discussion_r43962828

aroc: Ah ok, I think that's fair. We need `lodash` for the time being and if we're going to use it, we might as well make use of the convenience functions.

aroc:
Looks good! Let's merge!

PR #38, opened 2015-11-04
Updating data sent along with core events through Dispatcher

aroc:
It adds some additional data to the events being emitted through the AppDispatcher. This data is useful for event tracking and other actions that are performed in response to Explorer actions.

Perform some main actions in Explorer and notice the event data sent along is different and contains the new properties outlined in these changes.


josephwegner:
+1

PR #42, opened 2015-11-06
Fixing filter unpacking bug

aroc:
Fixes a bug where string filters can get unpacked as datetime filters.

Try creating a query with a filter that's a string, but with something that could get parsed by Javascript's date parser into a date, like: `query-string-1`.

Notice this filter gets properly turned into a `String` filter and not a `Datetime` filter.


aroc:
WIP: Just adding tests!

aroc:
Ok, ready for review.

josephwegner:
Merge it

PR #45, opened 2015-11-10
Fixes #44

### LINKED ISSUE #44: Remove incorrect query attributes when the analysis type changes
When the analysis type is changed from `count_unique` to `count`, the `target_property` does not get cleared from the query and remains both in the URL and in the data that gets sent to Keen when trying to save a query.

aroc:
This removes any invalid attributes from the query when it's updated. It looks at the analysis type and removes things like `target_property` for counts and extractions or `percentile` for non-percentile queries. This ensures that if a user makes a `count_unique` query with a `target_property`, then switches back to a normal `count` query then the `target_property` value gets cleared.

Try switching between query types and observe:
* That the URL updates and removes invalid query attributes
* That there are no errors when saving a query after changing around analysis_types and settings values for the additional fields required.

### Related Tickets
#44


joannecheng:
:100:  looks good to me

PR #46, opened 2015-11-10
QueryBuilder should place step parameters on the active step

josephwegner:
### What does this PR do?

This adds a bit of logic to the `QueryBuilder` component to know how/where to place parameters when updating the query. If it's a funnel, a certain set of properties need to be placed on the active step, while other properties can continue to be placed on the root query.

For the record, this is gonna get real messy if we ever start allowing timeframes on both the step and the root query. I'm gonna try not the think about that yet :)

### What makes you uneasy about this PR?

The function `placeUpdatedProperty` just feels... weird. Like, not the logic, but the fact that it exists, and _where_ it exists. I don't think there's a better place, and I think it's better to have that logic separated from `handleChanges`... it just still feels a tiny bit out of place.

### DIFF
--- a/client/js/app/components/explorer/query_builder/index.js
+++ b/client/js/app/components/explorer/query_builder/index.js
@@ -44,10 +44,10 @@ var QueryBuilder = React.createClass({
 
     if(_.isPlainObject(update)) {
       for(key in update) {
-        newModel.query[key] = update[key];
+        this.placeUpdatedProperty(newModel, key, update[key]);
       }
     } else {
-      newModel.query[update] = value;
+      this.placeUpdatedProperty(newModel, update, value);
     }
 
     ExplorerActions.update(this.props.model.id, newModel);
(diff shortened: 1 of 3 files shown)


aroc:
Ok, so I think you're 100% right that `placeUpdatedProperty` is not the way to handle this. While this absolutely would work for funnels and the code is relatively minimal, I don't believe it's the proper way to send updates to a specific step, for a few reasons:

1) The is too much logic for handling updates in a UI component. (I'm aware we have other components that potentially have similar amounts of this type of logic, but those should be refactored).
2) The update flow should be more direct and not require these conditional checks before updating. Rather than "send these updates, and if this is a funnel analysis type, then find the active step and update it instead of the main query", we should instead be using a  "dumber" and more direct path along the lines of "send these updates to the step in question".

So, how could we get an update path as direct as what I refer to in number 2? I think we could potentially solve this when we start working on the Funnels UI. There could be a "Funnel Manager" React Component that is passed **only** the funnels, and each funnel React component **only** gets passed a single step (which also has an ID, or we can even just use the index of the step in the steps array). That way, when a field is updated in a step component, the update path is dumb and direct.

If you take a look at the `FilterManager` and the `Filter` components, you can see this pattern at work (sort of). I built that a while ago and I've learnt a lot about React patterns since then, so we can likely do even better this time around.

I hope this makes sense. Let's chat more tomorrow and strategize. :)

josephwegner:
100% totally agree. You described my uneasiness perfectly.

Totally agree that a `funnelManager` or `stepManager` is the way to go with this. That seems pretty obvious, now that you say it :)

I think I could potentially just scrap this entire PR and handle that piece a bit later, when we start doing UI stuff.

josephwegner:
Closing, cause we have better ideas.

PR #48, opened 2015-11-11
Multi Group-by

aroc:
#### Note: Until Keen.js supports this, we cannot merge this PR.
This is now supported in Keen.js `3.4.0-rc2`. It will be a full release by the time we cut a new Explorer release - so this can be merged into master now.

It adds support in Explorer to optionally group by two properties. Our API supports this.

Try using the new multi-group by feature in the Query Builder section and see if the UI works as expected.

### Related Tickets
None.


joannecheng:
Minor nitpick: when I load a saved query with multiple group bys (or any group bys), the group by toggle doesn't open.

client/js/app/components/explorer/query_builder/group_by_field.js
```diff
@@ -24,25 +24,79 @@ var GroupByField = React.createClass({
...
+  },
+
   // React methods
 
   render: function() {
+    var secondGroupByField;
+    var toggleText = "Group by a second property";
+    var toggleIcon = "plus";
+
+    if (this.props.value.length > 1) {
```
joannecheng: how would you feel about pulling out the logic to generate the select fields and the multigroup toggle in a different private method?

client/js/app/components/explorer/query_builder/group_by_field.js
```diff
@@ -24,25 +24,79 @@ var GroupByField = React.createClass({
     return this.props.value;
   },
 
+  handleChange: function(name, value) {
+    var newVal = this.props.value.slice();
+    var index = name.split('.')[1];
+    newVal[index] = value;
+    this.props.handleChange(name.split('.')[0], newVal);
```
josephwegner: is `name.split('.')[0]` ever not `'group_by'`? Why not just hard code that?

client/js/app/components/explorer/query_builder/group_by_field.js
```diff
@@ -24,25 +24,79 @@ var GroupByField = React.createClass({
...
+  },
+
+  toggleMultiGroupBy: function(event) {
+    event.preventDefault();
+    var newVal;
+    switch (this.props.value.length) {
+      case 0:
```
josephwegner: when is this case hit?

client/js/app/components/explorer/query_builder/group_by_field.js
aroc: Yeah not a bad idea. Happy to do that! :)

client/js/app/components/explorer/query_builder/group_by_field.js
aroc: haha great eyes! You're absolutely correct here. Hard coding it is.

client/js/app/components/explorer/query_builder/group_by_field.js
aroc: The array is empty at first when no `group_by` has been set yet. So initial state basically.

aroc:
@joannecheng Not a minor nitpick, that's a bug. I'll look into it.

aroc:
Ok, this is ready for another review. @joannecheng Your issue with saved query loading has been fixed. The GroupByFields component has been refactored to make use of functions for some of the JSX to clean things up as well.

client/js/app/components/explorer/query_builder/group_by_field.js
josephwegner: +1 makes sense.

josephwegner:
+1

aroc:
Ok, this is ready for another quick review. @joannecheng The group by fields *do* open as expected now when loading from a saved query. :)

joannecheng:
:+1: Works for me now!

dustinlarimer:
:+1: all good here

aroc:
Merging!

PR #49, opened 2015-11-12
Refactor queryJSON and add funnel support

josephwegner:
### What does this PR do?

This refactors queryJSON, with the overall goal of adding funnel/step support.

The main item here is that, historically, we have never had to validate/format arrays of objects. Arrays of objects just kind of slipped by in `queryJSON`, and we hoped that it would get worked out later. Fixing that was the bulk of this PR - it adds some recursion to look at nested objects or nested arrays of objects. `FormatUtils.cleanQueryParameters` cleans them up, validates that their parameters actually belong in the query, and sends them back to `queryJSON`.

I think much of the logic that existed in `ExplorerUtils.queryJSON` actually belonged in `FormatUtils.cleanQueryParameters`, so I moved it there.

### What makes you uneasy about this PR?

Two things I'm worried about:

- The way I wrote this code is *not functional at all*. It relies on objects being passed by reference, and modifying the underlying object. I'm not returning new cloned objects - I'm just modifying the reference in place. I'm split on if this is cleaner or not - It would be quick to reformat this in a more functional approach, so would love some thoughts here.

- There is an existing pattern for filters that already kind of does what I wanted to achieve here... I could potentially create something like `StepUtils.queryJSON`, that works in a similar was as `FilterUtils.queryJSON`. This change would likely negate much of the work I've done, but I'm fine with that - better to spend the time architecting this correctly now, than to hate myself later.


josephwegner:
Oh, also, of the PRs I've done so far, this change is 100% by far the most likely to **break absolutely everything**. All the tests passed, and I've spent a good deal of time thinking about possible edge cased... I _think_ we're good, but this is a scary one.

I'm gonna hammer away on this PR running actual queries for a bit before I actually think it's mergeable.

aroc:
Ok, just took a look a this. My thoughts tend to be in line with your last two points in the PR description:

* We should try and keep all of our `Util` functions returning modified copies of whatever objects were passed in. Nearly (or all?) of our other Util functions have a return value and they don't modify underlying values. Things are much easier to reason about and easier to test if they follow a consistent pattern and don't have side-effects.
* I think you're also right about structuring things with a `StepUtils` (or maybe `FunnelUtils`) `queryJSON` function. I think having `ExplorerUtils.queryJSON` call additional `queryJSON` functions in more specific Util modules is an easy pattern to follow. It also means we can avoid having to do recursion and deal with concatenating strings together to check if the parameter is an allowed/valid one. If we had *many* arrays of nested objects, I'd say the recursion approach is likely the better one. But we only have filters and funnel steps that require this pattern with no additional ones in the foreseeable future, so having their own `queryJSON` functions isn't a ton of additional code and is easier to maintain.

Couple other small things, not from new code you've written but I noticed it reviewing this area of the codebase:
* Because we're using Util functions with return values, we can refactor things like this:
```
params.filters = _.map(params.filters, function(filter){
  return FilterUtils.queryJSON(filter);
});
```
Into this:
```
params.filters = _.map(params.filters, FilterUtils.queryJSON);
```

* I notice we're repeating the pattern of checking for absolute timeframes and removing the timezone key. We can move this into a FormatUtils function and share it across all the queryJSON functions.

josephwegner:
Closing, cause this one is so much cooler: https://github.com/keen/explorer/pull/57

PR #51, opened 2015-11-12
Fixes #50

### LINKED ISSUE #50: Datetime filters do not work (shows invalid date)
If you try to create a Datetime filter, the inputs show "invalid date" and you cannot edit them.

aroc:
It fixes #50.

Along the way, it refactors the handling of datetime filters. Specifically, it:
- Ensures all datetimes are managed in the local timezone up until the point where a query is made, a query is saved or the params are put into the URL. At that point, we strip the local timezone and add the timezone that's been chosen in the time range picker.

Try out Datetime filters. Try loading one from the URL. Notice that they behave as expected.

### Related Tickets
#50


josephwegner:
+1

PR #53, opened 2015-11-13
run queries automatically on pageload

### LINKED ISSUE #52: Queries do not automatically run on pageload
It used to be the case that queries would automatically run if you opened an Explorer page that had query attributes in the URL. This appears to have stopped happening.

The fix for this is likely simple, but I'm also going to see if I can get a unit test in for this. That may be more involved, since this happens on app initialization.

**Internal Ticket:** https://trello.com/c/Kg89AbeM/123-explorer-not-auto-loading-queries-from-url-on-pageload

josephwegner:
closes #52 

Looks like we had a mistaken `!` before `isEmailExtraction`, which was causing trouble. I removed that, and we're good to go.

I also set a sort of foundation for testing `App`.

### DIFF
--- a/client/js/app/app.js
+++ b/client/js/app/app.js
@@ -54,8 +54,9 @@ function App(config) {
   } else {
     AppStateActions.update({ ready: true });
     // Run the query for this explorer if it's valid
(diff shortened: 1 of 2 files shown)


joannecheng:
:100: :+1: :balloon:

PR #54, opened 2015-11-13
remove network dependency for unit tests

### LINKED ISSUE #47: Remove internet connection requirement for running tests
Tests make use of CDN hosted versions of jQuery and Keen.js. Ideally, you shouldn't need an internet connection to run tests.

josephwegner:
closes #47 

This moves keen-js and jquery dependencies into NPM, so that tests can load them without a network connection.


test/unit/index.js
```diff
@@ -1,5 +1,11 @@
 /** @jsx React.DOM */
 
+// **********************
+// Dependencies
+// **********************
+require('keen-js');
+window.$ = window.jQuery = require('jquery-browserify');
```
joannecheng: is this line needed? The tests seem to work with just `require('jquery-browserify')`.

josephwegner: You're probably right. I was just following their README :)

josephwegner:
cleaned up

joannecheng:
:+1:

PR #56, opened 2015-11-17
Funnels UI

aroc:
It adds the necessary React components to expose the `Funnels` analysis type. We've done some work at the data model layer to support this and much of those changes are already in `master`. This PR adds the Funnel step management layer as well as a few other updates at the data model layer, including changing how validations work depending on the analysis type (validating different things depending on whether we're dealing with a funnel query or not).

This PR also contains a pretty major refactor for the `FilterManager` and it's child components. The refactor made the `FilterManager` complete generic, such that it can be used many times in different parts of the code. The callbacks to handle filter additions, removal and changes are now passed in via `props` instead of living inside of the `FilterManager` itself.

Try out building a `funnel` query (by selecting the `funnel` analysis type). Check out that you can:
* Build and manage a funnel query in the UI, including adding steps, steps with filters, removing steps, etc.
* The Query can be saved
* The URL updates to reflect the funnel query attributes
* You can load a funnel query from parameters in the URL
* The API Query URL reflects the correct funnel query

### Tasks Remaining
- [x] Refactor validation to handle conditional validation sets better (we'll need this handle validation much different if the analysis type is `funnel`)
- [x] URL isn't updating immediately on model attribute change. It updates to the previous value on change (lagging behind by one change).
- [x] Disable caching for funnels until it's supported in the API
- [x] Chart type options not reflecting funnel analysis type
- [x] Better UI for displaying which steps have validation errors (rather than just global Explorer message)

### Bugs

- [x] Filters don't get applied
- [x] When adding a step, the new step is not set as active
- [x] When moving the top step up or the bottom step down, it is unset as active (eric)
- [x] Move analysis type about event collection by default
- [x] Removing a step should set a new step as active (joe)


josephwegner:
@aroc if it helps to pair down the PR size here, we could PR this into a dev branch `funnels`... that way we can review+merge this now, rather than waiting to be 100% ready to ship before this merges

aroc:
@josephwegner It's nearly complete, only validations updates and tests need to be added - probably easier to just add those and then get the whole thing reviewed. If I had the proper foresight, I would have done things like the FilterManager refactor as it's own PR. But alas, I did not. :/

josephwegner:
Can we move the analysis type selector _above_ event collection, for the default view? It feels weird to me that it jumps when switching to funnels

client/js/app/actions/ExplorerActions.js
```diff
@@ -3,10 +3,10 @@ var AppDispatcher = require('../dispatcher/AppDispatcher');
 var ExplorerConstants = require('../constants/ExplorerConstants');
 var ExplorerStore = require('../stores/ExplorerStore');
 var ExplorerValidations = require('../validations/ExplorerValidations');
-var ValidationUtils = require('../utils/ValidationUtils');
-var ExplorerUtils = require('../utils/ExplorerUtils');
 var NoticeActions = require('./NoticeActions');
 var AppStateActions = require('./AppStateActions');
+var RunValidations = require('../utils/RunValidations');
```
josephwegner: need to put the `.run` at the end of this require

aroc: We call it with `.run` below, so we're good. We have to do it this way for our tests to pass :/ Stubbing doesn't work with `require` calls that just return a function, as it would require stubbing out portions of the module loader which is craziness.

josephwegner: :D I commented in the wrong file:

https://github.com/keen/explorer/pull/56/files#diff-dca69ab628f9de688bfff269868227e4R6

josephwegner:
@aroc I wanted to call out this last commit 87e99de7d3245e4b6bbd955c0bfee9e1b1842025 as hacky and scary and something that we need to talk about.

The filters UI expects `isValid` to be on every filter property. This turns out to work pretty well when building a query, because filter validations seem to run on every filter as they are added. However, when loading a funnel query from the URL, we do not run validations on the specific filters. That means `isValid` doesn't exist.

So this fix just adds another nested loop that runs validations on every step filter, and applies the `isValid` value.

This exposes somewhat of an oddity of the current setup... the filter validations are actually run twice for every filter - once via `SharedValidations`, and once on each specific filter to apply the `isValid`. That's alotta validation. It also makes it hard to run validations on nested things, as we're seeing.

I'm not sure the best way to handle this. The way I did this is very not DRY, but I guess it works for now.

client/js/app/actions/ExplorerActions.js
aroc: Closing the loop: This has been fixed.

aroc:
Closing the loop: We're nearly good to go here. We're giving it some QA time then we're we'll merge. :)

PR #57, opened 2015-11-17
Add FunnelUtils to generate stepJSON

josephwegner:
### What does this PR do?

Three Things:

- Move all of the time-related functionality in `ExplorerUtils` to a new util called `TimeframeUtils`. This was mostly to solve circular references, and make my head stop exploding
- Clean up `ExplorerUtils.queryJSON` a little bit to be more succinct about how things run
- Add `FunnelUtils`, and call `FunnelUtils.stepJSON` from `ExplorerUtils.queryJSON`, so that we can actually send funnel queries!


client/js/app/utils/ExplorerUtils.js
```diff
@@ -79,36 +82,34 @@ module.exports = {
     }
     var params = _.cloneDeep(explorer.query);
 
-    // Set the timeframe (will get removed if it's null o undefined)
-    params.timeframe = module.exports.getTimeframe(explorer);
```
aroc: We set the timeframe on the explorer query to `null` when we move over to a `funnel` analysis type, so this function call will fail. We'll need `queryJSON` to take into account whether we're dealing with a `funnel` query or not and adjust what we try to parse/add to the params.

aroc: Whoops, meant for this to be down at the `TimeframeUtils.getTimeParameters` line. (line `85`).

aroc: We could potentially even just do a time `if (params.time)` without even really worrying out the specific analysis type.

aroc:
Ok this looks good other than the couple comments I made. I think we can merge after those updates. I'll need this merged for me to continue with Funnels UI work.

aroc:
Additionally, we're going to need `formatQueryParams` to be able to handle unpacking steps in the same way it does with filters. Likely easiest to keep the same pattern and add a `formatQueryParams` function to `FunnelUtils`.

client/js/app/utils/ExplorerUtils.js
josephwegner: @aroc if you pass in a null or undefined time value, you will get a null timeframe back... which then means it gets plucked by `FormatUtils.isValidQueryValue` later on.

josephwegner:
I'll do unpacking in a PR real quick.

client/js/app/utils/ExplorerUtils.js
aroc: Oh interesting, this was failing for me. I'll take another look.

aroc:
:100:

PR #60, opened 2015-11-20
a few fixes for loading in funnel steps on init

josephwegner:
A few init fixes here for loading funnels in at app initialization.

@aroc there is an issue here regarding nested forms in the funnel builder... 

`Uncaught Error: Invariant Violation: findComponentRoot(..., .0.1.0.0.1.0.2.1.$0.0.1.0): Unable to find element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID `

### DIFF
--- a/dist/keen-explorer.css
+++ b/dist/keen-explorer.css
@@ -1792,13 +1792,13 @@
   bottom: 6px;
   border-left: 5px solid transparent;
   border-right: 5px solid transparent;
-  border-bottom: 5px solid #FFF;
+  border-bottom: 5px solid #ffffff;
 }
 #keen-explorer .funnel-builder .step-move-btns .down .arrow {
   top: 6px;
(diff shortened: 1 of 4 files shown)


dist/keen-explorer.css
```diff
@@ -1792,13 +1792,13 @@
   bottom: 6px;
   border-left: 5px solid transparent;
   border-right: 5px solid transparent;
-  border-bottom: 5px solid #FFF;
```
aroc: hm, these are weird. Probably just weird artifacts from the build. ¯\_(ツ)_/¯

aroc:
Looks good!! :D thanks for this!

PR #61, opened 2015-11-20
Funnels UI:  Funnel visualization support

aroc:
This fixes the issue where we couldn't visualize funnel query responses. Keen.js requires the entire response in order to properly auto-visualize a Funnel query. Rather than just storing the result of a query in the model (ExplorerStore models), we're instead storing the entire response.


josephwegner:
tests are broke, but the logic here seems solid. +1 after tests.

aroc:
Fixing tests! :dancer:

PR #62, opened 2015-11-20
Funnels UI: Step event handling

aroc:
This fixes the issue where clicking a "move step" button would also call the "toggleStepActive" callback.


aroc:
Whoops, wrong branch to merge into!

PR #64, opened 2015-11-23
Funnels UI: Nested validations updates and more test coverage

aroc:
This PR addresses two major outstanding issues with the main Funnels UI PR (#56):
* Nested validations are a pain and require nested loops with lots of duplicated code in the `ExplorerStore` `_validate` function.
* There are not enough tests covering the new Funnels related functionality.

### The fix for nested validations
Nested validations are much cleaner now because we're setting the validation information, specifically the `isValid` boolean property and the `errors` array, directly on each object to be validated **when it is validated**. The `RunValidations.run` function no longer returns the results of the validation, but rather changes the underlying object. While this approach is different than the rest o the Utility modules, I think it makes sense here given how we're using validations in the application - and the significantly reduced code here reflects that.

### More tests
More tests have ben added for the `FunnelBuilder` component as well as the `FunnelStep` component. They are by no means exhaustive, but they add some coverage for core interactions and expectations for how the components should behave.


aroc:
I've just recently added support for showing notices for validation errors at the step level. The way I've made them obvious to the user is that I make the first step with a validation error active so that it opens up and displays the message. We could probably do more here, but it's pretty obvious looking to me. Let me know what you all think of it:

image

client/js/app/components/explorer/query_builder/funnels/funnel_step.js
```diff
@@ -76,9 +79,9 @@ var FunnelStep = React.createClass({
   },
 
   buildStepBody: function() {
-    var remove;
+    var remove, notice;
 
-    if(this.props.canRemove) {
+    if (this.props.canRemove) {
```
josephwegner: :facepunch: I'm an idiot.

client/js/app/utils/ExplorerUtils.js
```diff
@@ -65,7 +69,7 @@ module.exports = {
   },
 
   shouldHaveTarget: function(explorer) {
-    return ANALYSIS_TYPES_WITHOUT_TARGET.indexOf(explorer.query.analysis_type) === -1;
+    return !FormatUtils.isNullOrUndefined(explorer.query.analysis_type) &&ANALYSIS_TYPES_WITHOUT_TARGET.indexOf(explorer.query.analysis_type) === -1;
```
josephwegner: space after `&&`

client/js/app/utils/RunValidations.js
```diff
@@ -1,23 +1,28 @@
...
       if (result === true) continue;
-      var errorMsg = (result === false) ? validator.msg : result;
-      errors.push({
+
+      // There is an error. Set the model as invalid and set the errors on the model.
+      model.isValid = false;
+      model.errors.push({
         attribute: keys[i],
-        msg: errorMsg
+        msg: (result === false) ? validator.msg : result
```
josephwegner: This will always set to `validator.msg`, won't it? Because of the `if (result === true) continue;` line above?

client/js/app/components/explorer/query_builder/funnels/funnel_step.js
aroc: :P

client/js/app/utils/ExplorerUtils.js
aroc: Whoops, good eye!

client/js/app/utils/RunValidations.js
aroc: Nope, it can also be a string. If so, then that string should be the validation message. This is basically a crappy hack to allow a validator to return different error messages depending on the "type" of validation failure. See the `time` validator to see how it works in action.

josephwegner: dur.

Makes sense

josephwegner:
+1 after that `&&` space

PR #69, opened 2015-12-04
Gzipping for explorer!

aroc:
It adds gzipping the javascript to the explorer build process, reducing the size of the build file by 74% down to only 152kb gzipped.

Try the `production` gulp task - ensure the gzipping happens as expected.


joannecheng:
Everything works on my end :sheep:

PR #71, opened 2015-12-15
Allow saving of Funnel queries

aroc:

It removes the flag that prevent users from saving funnel queries. While caching funnel queries is not yet supported by the API, saving them is.

Try creating and saving a funnel query.

- [x] Does the UI, pixel by pixel, look exactly as expected?
- [x] If the feature makes requests from the browser, inspect them in the Web Inspector, do they look as expected (parameters, headers, etc)?
- [x] If the feature sends data to Keen, is the data visible in the project if you run an extraction?


sfalkoff:
@aroc, hmm I got this error trying to save (not cache) a one step funnel query (http://localhost:8081/?query%5Banalysis_type%5D=funnel&query%5Bsteps%5D%5B0%5D%5Bevent_collection%5D=create_organization&query%5Bsteps%5D%5B0%5D%5Bactor_property%5D=user.uuid&query%5Bsteps%5D%5B0%5D%5Btimezone%5D=UTC&query%5Bsteps%5D%5B0%5D%5Boptional%5D=false&query%5Bsteps%5D%5B0%5D%5Binverted%5D=false&query%5Bsteps%5D%5B0%5D%5Btimeframe%5D=this_14_days): 
screen shot 2015-12-15 at 3 33 20 pm

I'm seeing 6 pending checks after running `npm run production` -- is that expected?

aroc:
@sfalkoff Interesting. I'll have to try it again. Maybe I didn't actually try to save (totally thought I did). Maybe it's not actually supported by the API yet. Lemme check with Terry.

aroc:
Ok, looks like it *is* supported. I'm gonna run a couple more tests on my end.

aroc:
Whoops, MID got back to me and corrected: It is *not* supported yet. Gonna have to wait on this one.

aroc:
Ok, back merged master into this branch so it's caught up. We now have a bug though: The chart doesn't render with latest master merged in, likely as a result of how we're handling chart type determination inside `ExplorerUtils.getChartTypeOptions`. Will need to update to support funnels.

aroc:
Fixed the issue preventing funnels from being properly visualized. This is ready to merge baby merge!

PR #75, opened 2015-12-18
Fix errors thrown when metadata is not present or group_by field is not an array

aroc:

It ensures all new models created in the `ExplorerStore` are created with a `metadata` property that has the expected structure. This prevents older saved queries from breaking the explorer when this data that is expected is not present.

The same issue was happening for the `query.group_by` field on the models. Older saved queries have the `query.group_by` field as a string, not an array. But the new React group_by_field component requires an array to work correctly. So `ExplorerStore` 's `_create` function has been updated to turn string fields into arrays.

## How should this be tested?

- [x] Do all tests pass?

## Related tickets?
N/A


aroc:
Merging this in - it fixes a critical bug that it's master.

Issue #77, opened 2015-12-21
Saving a query breaks explorer: Cannot read property 'length' of null


wetzler:
I tried to create a simple query for testing purposes. Saving it seems to break the explorer.

Steps to repro:
1. Go to the project.
2. Select a count for the collection "data"
3. Enter timeframe "this 14 years"
4. Run Query.
5. Enter title "Simple Query 2"
6. Click Save query.

Result => Error in console: "cannot read property 'length' of null". 
The explorer sort of breaks and I can't even create new queries. 

I will send you the link to try it out  yourself.

aroc:
Fixed by #79

PR #79, opened 2015-12-21
Fix for #77

aroc:
Updating ExplorerStore models did not have the right logic to wrap incoming group_by properties in an array.

## How should this be tested?

Try saving a query and seeing if you can load it back up. Try one *with* and one *without* a group by option.

- [x] Do all tests pass?

## Related tickets?
#77


joannecheng:
Looks good! Do you also need to update `keen-explorer.min.js`?

aroc:
Ah yes, I do! Good catch!

aroc:
Ok, build files committed. Merging.

PR #84, opened 2015-12-22
Fixes #81

### LINKED ISSUE #81: “group_by” param in the API Query URL does not group
There is an issue w/ the “group_by” param in the API Query URL that you generate in the Explorer that, in effect, *does not* group the query results :

api.keen.io/3.0/projects/project/queries/count?api_key=key&timezone=US%2FEastern&timeframe=this_30_days&**group_by%5B0%5D=app.name**&event_collection=Viewed%20photo&filters=%5B%5D

I bolded the text — but here is what it does: &group_by%5B0%5D=

Realized this an hour into futzing with it and trying to figure out why I couldn’t import the JSON into a Google Doc. Removing that invalid character fixed it immediately.

Am giving you a heads up, since I assume this is a widespread issue and not unique to our account.

aroc:
Ensures group_by property of queries is encoded into JSON before being uri encoded - as per the Keen API. We were not doing this before because we were not supporting multi-group by - but now that we are, we must do this.

## How should this be tested?

1) Try creating single and multi-group by queries.
2) Run them.
3) Copy and paste the API Query URL into a new tab.
4) Ensure the response JSON looks as expected.

- [x] Do all tests pass?

## Related tickets?
#81


client/js/app/utils/ExplorerUtils.js
```diff
@@ -265,6 +265,10 @@ module.exports = {
       delete attrs['steps'];
     }
 
+    if (attrs.group_by) {
```
joannecheng: is this `if` statement needed? `JSON.stringify(undefined)` is still undefined, and it doesn't look like `group_by` shows up in the URL when `attrs.group_by` is `undefined`.

joannecheng: Plus, all the tests still pass and everything works as expected

aroc: Hmmmm, actually I think this may be a problem. We only want to stringify the group_by property when it's an array, otherwise we just want to encodeURIComponent without stringifying it. So this should be: `if (attrs.group_by && _.isArray(attrs.group_by) && attrs.group_by.length > 1)`

We could potentially send the API a single item array stringified, but it feels a bit weird to have people using API urls that have single item arrays stringified. It might mislead them into thinking it *has* to be that way, when in fact for a single group by you can just do standard url encoding.

Does that make sense? I realize this all might be confusing haha.

joannecheng: Yep, I get it!

aroc: Ok cool, I'll make that change now and write another test to cover the array and non-array use cases.

aroc:
@joannecheng ok, ready for another quick look.

joannecheng:
:+1:

PR #86, opened 2015-12-22
Fixes #85

### LINKED ISSUE #85: explorer doesn't properly handle queries with result: 0
Queries which return the API response of {result: 0} are incorrectly displayed in the explorer as "Your query returned no results". 

That might make sense on some queries e.g. Count the number of events where myproperty = "feather monkey brain", but it doesn't make sense on mathematical operations where the result is 0. 

For example, a user is running a percentile query on a dataset where the value is often zero. For percentiles less than 38, they should get "0" as the result. Instead they get "Your query returned no results". 

Another point is that there is no way to toggle to see the JSON response from the query. If you could, it would be easier troubleshoot and see the result of 0.

screen shot 2015-12-22 at 1 05 59 pm

aroc:
Ensures results with a value of 0 properly show a metric.

## How should this be tested?

Try running a query that you know for sure will return a result of 0. Here's a good example:

1) Run a query for an event that happens infrequently and set the timeframe to `this_1_minutes`.
2) Observe that the chart area shows a metric with a value of 0 instead of a message of "no results".

- [x] Do all tests pass?

## Related tickets?
#85


dustinlarimer:
:+1: Looks ok to me

PR #87, opened 2016-01-04
Add 'repository' field to package.json

joannecheng:

This removes the warning "keen-explorer 2.0.0: No repository field"

## How should this be tested?

Run `npm install`. The warning message should not be at the end of the output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,10 @@
       "reactify"
     ]
   },
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:keen/explorer.git"
+  },
   "devDependencies": {
     "browserify": "11.2.0",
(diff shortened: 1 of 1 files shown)


sfalkoff:
@joannecheng I tested this out and do not see the warning message addressed here. 

I'm seeing two other warning messages after running `npm install`, however I'm also seeing them on master -- perhaps those are existing issues that can be addressed separately. Here are the messages -- are you seeing these as well? 

screen shot 2016-01-05 at 10 01 29 am

joannecheng:
@sfalkoff I am seeing them, but they should be addressed separately and not in this PR.

sfalkoff:
@joannecheng alrighty, looks good. :shipit:

PR #90, opened 2016-01-08
Fix for saved queries with null start or end absolute timeframe values

aroc:
It fixes a bug that causes Explorer to crash if a saved query that gets loaded has an absolute timeframe with a null start or end time.

## How should this be tested?

Try loading a project with saved queries that have absolute timeframes where either the start or end properties are null.

- [x] Do all tests pass?

## Related tickets?
N/A


sfalkoff:
@aroc looks good! :ship:

PR #92, opened 2016-01-08
Performance Update

aroc:
It fixes a couple major outstanding issues that directly affected performance.

1) It ensures the chart is not rendered on every component update and instead only re-renders the chart if there has been a new query run. This is handled by timestamping the query after the response comes back from the server and comparing it against the timestamp of the last rendered visualization. If they do not match, a render happens. Also, a changing of chart type can also cause a render.

2) It no longer sorts event collection properties each time a new collection is chosen. With larger projects, sorting a property array can be *highly* CPU intensive and causes massive dropped frames during the sorting. This is the cause of the perceived freezing of Explorer when changing to an event collection that has a large number of properties or changing other UI elements when one of those collections is already selected.

The new approach is to instead take additional time at initial Explorer load to loop through all event collections and sort each property list once so they're all pre-sorted ahead of use. This frees up significant CPU cycles and ensuring Explorer continues to run smoothly when switching between event collections or interaction with a very large event collection (large in terms of number of properties).

## How should this be tested?

Testing point 1 above:
1) Run a query.
2) Interact with the UI and notice the chart does not re-render. This will require breakpoints or debuggers inside the codebase to ensure it works as expected.
3) Change the chart type and ensure it works as expected.
3) Repeat several times with various queries and chart types.

Testing point 2 above:
1) Use a project with a large schema, specifically one with event collections that you know have a very large number of properties.
2) Use the UI with these collections and ensure things work as expected.

- [x] Do all tests pass?

## Related tickets?
N/A


sfalkoff:
Reviewed. :ship:

aroc:
Awesome! Thanks!

PR #93, opened 2016-01-13
Fix for group by and interval toggles

aroc:

It fixes a bug where query builder fields that are hide-able with the `FieldsToggle` component were not opening when a saved query was loaded. This was because of how the initial open state is determined. This has been updated and now saved queries loaded up that have values for fields hidden behind toggle-able fields are shown as open at load time.

## How should this be tested?

1) Load a saved query with interval or group by fields. Confirm they behave as expected: they are initially open.
2) Try playing with the form fields behind the `FieldsToggle` components and ensure they behave as expected: they open and close as expected and are in the open state and closed state when expected.

- [x] Do all tests pass?

## Related tickets?
N/A


aroc:
Shipping this without external review, as it directly affects someone using this project in production.

PR #98, opened 2016-01-15
Performance Updates

aroc:
It makes some changes to the codebase to increase performance, specifically when dealing with large responses / large query result datasets.

The changes it makes:
- No longer using `_.cloneDeep` everywhere we make updates. This cloning is very costly. Instead, we only pass in the properties we need to update and use a `_.merge` call. We also *never* clone the response object and instead just assign the new model's `response` reference to the old model's `response` object.
- We no longer initialize a new `Dataviz` object with the entire response object each time we want to check which chart types are supported for the given query. We borrowed some code from Keen.js and put it into this project for *much* faster checking of what chart types are supported.

## How should this be tested?

1) Try running some queries with large results, especially ones with a large number of groups.

- [x] Do all tests pass?

## Related tickets?
N/A


client/js/app/actions/ExplorerActions.js
```diff
@@ -217,7 +217,8 @@ var ExplorerActions = {
     updates.loading = false;
 
     if (!ExplorerUtils.responseSupportsChartType(response, explorer.metadata.visualization.chart_type, explorer.query.analysis_type)) {
-      updates.metadata.visualization.chart_type = ExplorerUtils.getChartTypeOptions(response, explorer.query.analysis_type)[0];
+      updates.metadata = _.cloneDeep(explorer.metadata);
+      updates.metadata.visualization.chart_type = ExplorerUtils.getChartTypeOptions(response.query)[0];
```
joannecheng: Getting this error when I run a `count` query: `Uncaught TypeError: Cannot read property '0' of undefined`

joannecheng: project id: 5011efa95f546f2ce2000000
count on `pageviews`

client/js/app/utils/ExplorerUtils.js
```diff
@@ -210,32 +207,74 @@ module.exports = {
...
+    if (query.analysis_type === "select_unique") {
+      dataType = 'nominal';
+    }
+
+    return dataType;
+  },
+
+  getChartTypeOptions: function(query) {
+    var dataTypes = {
+      'categorical':        ['piechart', 'barchart', 'columnchart', 'table'],
```
joannecheng: it looks like `singular` is missing from the `dataTypes` object in `getChartTypeOptions` in ExporerUtils

joannecheng: adding `'singular':           ['metric']` to the object seems to fix it

client/js/app/actions/ExplorerActions.js
aroc: Looks like the `singular` data type addition fixed this. Can you take another look when you get a moment?

joannecheng:
:+1: everything looks good now.

PR #100, opened 2016-01-21
Fix group-by fields and Timeframe type toggle (PLUS: lodash upgrade)

aroc:

It fixes two bugs in Explorer:

1) When adding a second group by option, you could not remove that second group by field after adding it. This was due to how updates are merged into the model, a change that was introduced in #98.
2) When trying to toggle between absolute and relative timeframes, it would not work. You'd get "stuck" on an absolute timeframe after selecting it. This was due to how updates are merged into the model, a change that was introduced in #98.

There are two new tests added to cover these two scenarios.

#### PLUS: Upgrade to lodash 4.0.0
This PR also upgrades to latest lodash because we needed the `_.mergeWith` function to fix these bugs. :palm_tree: 

## How should this be tested?

1) Try using the group by fields in the query builder. Ensure they work as expected.
2) Try toggling between absolute and relative timeframes. Ensure they work as expected.

- [ ] Do all tests pass?


client/js/app/stores/ExplorerStore.js
```diff
@@ -128,7 +128,14 @@ function _getDefaultFilterCoercionType(explorer, filter) {
...
   // don't have to loop through the (sometimes) massive response object.
-  var newModel = _.merge({}, explorer, _.omit(updates, 'response'));
+  function customizer(objValue, srcValue, key, object, source) {
+    if (_.isArray(objValue)) {
+      return srcValue;
+    } else if (key === 'time' && _.isPlainObject(objValue)) {
+      return srcValue;
+    }
+  }
+  var newModel = _.mergeWith({}, explorer, _.omit(updates, 'response'), customizer);
```
joannecheng: I'm getting an error that `mergeWith` is not a function with the latest version of lodash.

client/js/app/stores/ExplorerStore.js
```diff
@@ -128,7 +128,14 @@ function _getDefaultFilterCoercionType(explorer, filter) {
 function _prepareUpdates(explorer, updates) {
   // TODO: We're assigning the response object directly onto the model so we
   // don't have to loop through the (sometimes) massive response object.
-  var newModel = _.merge({}, explorer, _.omit(updates, 'response'));
+  function customizer(objValue, srcValue, key, object, source) {
```
joannecheng: `object` and `source` aren't used anywhere in the function. It might be a bit cleaner to remove them from the argument list.

test/unit/stores/ExplorerStoreSpec.js
```diff
@@ -221,6 +221,62 @@ describe('stores/ExplorerStore', function() {
...
+      var explorer = ExplorerStore.get('SOME_ID');
+      var updates = { query: { group_by: ['name'] } }
+      ExplorerActions.update('SOME_ID', updates);
+
+      explorer = ExplorerStore.get('SOME_ID');
+      
+      assert.sameMembers(explorer.query.group_by, ['name']);
+    });
+
+    it('should properly the time object', function () {
```
joannecheng: should properly what the time object? :)

joannecheng:
Just a few comments, but otherwise everything works as it should. :confetti_ball:

client/js/app/stores/ExplorerStore.js
aroc: Agreed!

test/unit/stores/ExplorerStoreSpec.js
aroc: lewl! good catch

aroc:
Cool! Fixed up. Merging.

PR #101, opened 2016-01-21
Refactor ExplorerUtils, Part 1

joannecheng:
## What does this PR do?

ExplorerUtils is a 428 line file, and its spec file is 628 lines. This makes ExplorerUtils difficult to navigate and maintain. There are also several functions in ExplorerUtils that are only used once in the application and belong in other places.

This PR removes some functions from ExplorerUtils into modules, appropriate components, or out of the application entirely if they're not being used anywhere. This PR is meant to be a first step in cleaning up ExplorerUtils so it's more maintainable and only contains functions that are used in multiple places in Explorer.

Additionally, this PR cleans up `visualization/chart.js` and its spec file so it conforms to the app's style.
## How should this be tested?

- [ ] Do all tests pass?


client/js/app/modules/api_query_url.js
```diff
@@ -0,0 +1,71 @@
+var _ = require('lodash');
+var Qs = require('qs');
+var ExplorerUtils = require('../utils/ExplorerUtils');
+var TimeframeUtils = require('../utils/TimeframeUtils');
+
+function ApiQueryUrl(client, explorer) {
```
aroc: What was the rationale for using a constructor here? It looks like there's only one function used (`getUrl`) and no need to keep object properties as state between function calls. You could export a single function here and not require the use of `new` when using the function in the `<QueryBuilder>` component.

client/js/app/components/explorer/visualization/chart.js
```diff
@@ -11,88 +11,102 @@ var FormatUtils = require('../../../utils/FormatUtils');
...
-	  	});
-	  	chartContent = (
-	  		<textarea ref='jsonViz' className="json-view" value={content} readOnly />
-	  	);
-		}
-		else {
-	  	chartContent = (
-	  		<KeenViz model={this.props.model} dataviz={this.props.dataviz} />
-	  	);
-	  }
```
aroc: Thanks for fixing this brutal indentation! very likely something I committed :P

client/js/app/components/explorer/visualization/chart.js
```diff
@@ -11,88 +11,102 @@ var FormatUtils = require('../../../utils/FormatUtils');
...
+        <div className="alert alert-danger">
+          <span className="icon glyphicon glyphicon-info-sign error"></span>
+          {'Your query returned no results.'}
+        </div>
+      </div>
+     );
+
+  },
+
+  isJSONViz: function() {
```
aroc: Cool, these functions all make much more sense to have here. :+1:

client/js/app/components/explorer/query_builder/index.js
```diff
@@ -199,7 +200,7 @@ var QueryBuilder = React.createClass({
   render: function() {
     var apiQueryUrl;
     if (this.props.model.isValid) {
-      apiQueryUrl = ExplorerUtils.getApiQueryUrl(this.props.client, this.props.model);
+      apiQueryUrl = new ApiQueryUrl(this.props.client, this.props.model).getUrl();
```
aroc: Regarding my comment below about the ApiQueryUrl prototype: If you exported a single function this would become something along the lines of: `apiQueryUrl = GetApiQueryUrl(this.props.client, this.props.model);`

aroc:
Ok, done main review! Everything looks great! Nice refactor.

aroc:
Whoops, looks like I'm hitting an error when running a query with a group-by:
image

Query URL:
http://localhost:8081/?query%5Bevent_collection%5D=add_project_member&query%5Banalysis_type%5D=count&query%5Bgroup_by%5D%5B0%5D=project.id&query%5Btimezone%5D=UTC&query%5Btimeframe%5D=this_14_days

`demo/index.html` is using keen.js version `3.4.0-rc3`.

client/js/app/modules/api_query_url.js
joannecheng: I think I had grand schemes of breaking out the whole function into smaller private functions, which I guess is possible without creating a new object. You're right, I can just export one function and avoid creating a new object.

joannecheng: Would this still belong in `modules`, then? It almost feels like it would belong in `components/explorer/query_builder`, except everything else in that file is a react component.

joannecheng:
@aroc I wasn't able to replicate that group by error - it might have been fixed in the commits from the past few days. Could you check again? I'm also not sure where to put `api_query_url.js`.

aroc:
Closing this PR as it's very old and we have no current intention of updating it.

PR #102, opened 2016-01-21
Fixes infinite loop when rendering chart

aroc:

It fixes a bug where Explorer would get into an infinite loop when rendering the chart. This was because we were using state to keep track of the lastDataTimestamp, which caused re-renders. Instead, we're just using a component property.

## How should this be tested?

Try loading different saved queries one after the other and ensure they do not cause an infinite loop inside of `showVisualization` inside of the `KeenViz` component.

- [x] Do all tests pass?


client/js/app/components/explorer/visualization/keen_viz.js
```diff
@@ -7,11 +7,14 @@ var React = require('react');
...
 var KeenViz = React.createClass({
 
+  lastDataTimestamp: null,
+
 	// ***********************
 	// Convenience functions
 	// ***********************
 
 	showVisualization: function() {
+    console.log(new Date(), 'INSIDE showVisualization!')
```
joannecheng: Accidental `console.log` statement?

joannecheng:
I don't see an infinite loop anymore when the chart renders, good job! Just need to remove the extra console.log statments.

client/js/app/components/explorer/visualization/keen_viz.js
aroc: Nooooooo!!!!!

PR #103, opened 2016-01-22
Fix visualization selector

joannecheng:

This fixes a few bugs:

* User is not able to run a query of a different data type without visualizations breaking.
* User is not able to switch between different chart types.

## How should this be tested?

* Run a count query in explorer. Add an interval, re-run the query. Observe that a chart gets rendered.
* Run a query in explorer that has multiple chart types (example: interval of a count). Switch between the different types of charts many times. Observe that the chart type changes to the appropriate visualization.

- [x] Do all tests pass?


aroc:
Can't seem to properly change between chart types still. :/ Here's a video: http://d.pr/v/1dRe1

It seems to work when changing FROM json TO new chart type. But not FROM  a chart type (not json) TO another chart type.

aroc:
Whoops nevermind, on the wrong branch. sorry! Looks great! :ship:

PR #104, opened 2016-01-23
WIP: Replace keen js with keen dataviz

joannecheng:
Note: To run, you need to load `keen-dataviz.js` and `keen-dataviz.css` in `demo/index.html`


dustinlarimer:
We could load `keen-dataviz` over npm to keep it all tidy.

dustinlarimer:
Now handled in #133 :+1:

PR #111, opened 2016-01-29
Fixes #65: Boolean filters cannot be loaded from URL

### LINKED ISSUE #65: Boolean filter values handled as strings
Filter values of `true|false` fail internal validation, resulting in an error message of:

> The specified value is invalid for the given operator. Operator: exists. Value: false

aroc:

It fixes a bug where the `exists` filter could not be properly loaded from the URL. This was because the filter is being treated as a `string` type filter instead of `boolean`. This fixes that bug and also adds test coverage for it.

## How should this be tested?

- [x] Do all tests pass?

## Related tickets?
#65


test/unit/utils/FilterUtilsSpec.js
```diff
@@ -22,6 +23,16 @@ describe('utils/FilterUtils', function() {
         };
         assert.strictEqual(FilterUtils.getCoercionType(filter), "String");
       });
+
+      it('String type when property_value can be parsed into a date', function () {
```
sfalkoff: Perhaps this test should be described as `'Boolean type when filter exists'` ? ... or something like that

sfalkoff:
@aroc thanks for fixing this. +1 after comment

test/unit/utils/FilterUtilsSpec.js
aroc: Whoops, good eye! I copied the previous test and didn't update the name.

adambraimbridge:
Any chance of a patch release for this? It missed the 2.2.0 release in Jan.
https://github.com/keen/explorer/compare/v2.2.0...master

Issue #110, opened 2016-01-27
"Can't get my saved query to work" (explorer re-named it in all lowercase)


wetzler:
I want to setup a dashboard that uses a saved query. So, I saved a new one in the Explorer and now I'm following the js-guide for how to embed it in my dashboard. However, I keep getting an error that the resource does not exist!? I'm being super careful, double-checking everything, even copy & pasting the name of the saved query from the explorer so that I get it exactly right. I'm trying every API key and variation on a saved query URL that I find in the API reference. What am I doing wrong!? Do saved queries even work!? Is this API reference incorrect? What is the deal here!? Finally, I'm inspecting the explorer to try to see what network request is actually being sent when it loads the chart.... ahh.....

After a lottttt of troubleshooting by no less than 3 Keen folks, we finally realized that saved query URL is case sensitive, and that the explorer had named the query in all lowercase.

Steps to repro:
1. In the explorer, create a saved query and name it "myTshirtRequests"
2. Now try to embed a saved query with the name "myTshirtRequests"
(it only works if you rename it to "mytshirtrequests")

Turns out there is a lower-case version of this query displayed here, but it's not clear how/why that is there or that I should use that when embedding saved queries.

![screen shot 2016-01-27 at 10 14 39 am](https://cloud.githubusercontent.com/assets/1840448/12623547/84dff7fe-c4df-11e5-9149-ab1676bb9a20.png)

aroc:
Started discussions for finding a better approach with the team. Thoughts:
1) Do we *need* the display name version of the saved query? Maybe the actual query name that's used to get the saved query is the only field needed? Although this does mean that the Explorer list of queries would no longer have these "nice name" versions which are allowed to have spaces, capitalized letters etc.
2) What would the indication that the bottom field is the "query name to be used when dealing with the API" look like? A little question mark icon that when clicked reveals more information?

wetzler:
My personal opinion:
Do away with "display name". It's not THAT much more useful to have MyTshirtQuery vs my_tshirt_query. That being said, I'm a developer so I'm more comfortable with more technical-looking labels.

dustinlarimer:
Great suggestions - we'll discuss further this and all the implications for how these queries get used.

dustinlarimer:
I'll rework the visual aspect of this as-is and see if we can find a more informative and intuitive approach first.

aroc:
Closing as we've merged a redesign of this component.

PR #113, opened 2016-02-01
Update saved query name input style to address #110

dustinlarimer:

This PR addresses #110 by clarifying the saved query name component. State is now tracked to hide/show the query name input, and a label has been added to clarify that the second input field is the saved query resource that one would access via the API.

Ready state:
screen shot 2016-02-01 at 3 47 06 pm

Active state:
screen shot 2016-02-01 at 3 47 02 pm

## How should this be tested?

Compose a new query and interact with the Display Name/Title component

_Fetch the branch and/or deploy to staging to test the following:_

- [x] Do all tests pass?

## Related tickets?

#110


joannecheng:
Awesome improvements! Just make sure to run `gulp scripts` and `gulp styles` so update the minified and compressed scripts and styles.

aroc:
You can produce all the production build files with `npm run production`.

dustinlarimer:
@joannecheng @aroc got it, thanks!

Issue #29, opened 2015-10-05
Allow filters to equate to an empty string.


hex337:
Right now, you can't create a filter to check for the equalness or non-equalness of an empty string (it just removes the filter). There are cases where you would want to check for an empty string or when that field is not empty.

dustinlarimer:
:+1: we'll look into this, thanks!

SteeleParker:
Just a little further info:

Currently the API allows the checking of `equal to` or `not equal to` an empty string, however the explorer gives an error when attempting to save a filter with either of these constraints

For Example, the following is an accepted filter through the API
`{property_name: "myProperty", operator: "ne", property_value: ""}`

PR #114, opened 2016-02-02
Allow empty strings as filter property values

dustinlarimer:

This PR addresses #29 by allowing empty strings as valid filter property values.

## How should this be tested?

* `/utils/FormatUtils`: `isList` updated to reject cases where no match is found
* `/validations/FilterValidations`: allow `coercionType === 'String'` to pass

- [x] Do all tests pass?

## Related tickets?

#29


client/js/app/validations/FilterValidations.js
```diff
@@ -46,7 +46,7 @@ module.exports = {
         return true;
       } else if (coercionType === 'Number') {
         return _.isNumber(value);
-      } else if (String(value) === "0") {
+      } else if (coercionType === 'String' || String(value) === "0") {
```
aroc: You no longer need the `String(value) === "0"` check now that we're allowing empty strings.

dustinlarimer: Cool, removed this in my latest commit :+1:

PR #122, opened 2016-02-02
Fixes #112: Missing table view option

### LINKED ISSUE #112: Table view is missing
As a user, I want to be able to easily copy & paste the results of my aggregations for use in excel, google sheets, keynote, etc.

I used to use the "table" visualization type for this, but recently I noticed it is missing from the explorer.

aroc:
This ensures the proper data type is being set and that the table option is available when it should be.

## How should this be tested?

Try running a query with an interval and ensuring that the `table` view type is available to be used and works as expected.

- [x] Do all tests pass?

## Related tickets?
#112


sfalkoff:
Looks good! Merging.

PR #126, opened 2016-02-12
Removed old copy

tbarn:
Corrects some copy in the Explorer that shows when doing an extraction. Keen IO API now supports saved queries. 

## How should this be tested?
Visually check that the copy is now accurate. Try to do an extraction with the "Bulk CSV extraction by email" radio button to see where it appears in the Explorer. 

## Related tickets?
Couldn't find any


dustinlarimer:
Thanks @tbarn :raised_hands:

PR #130, opened 2016-03-24
Avoid overwriting the React global variable

stof:
I don't see a need to expose the React when loading the explorer.

I assume that this was done initially to make it possible to use the React DevTools, but the integration between React and its DevTools was changed in React 0.12 to avoid requiring to have React as a global (instead, the DevTools are exposing a global variable)

### DIFF
--- a/client/js/app/app.js
+++ b/client/js/app/app.js
@@ -96,7 +96,6 @@ App.prototype.render = function() {
   }), this.targetNode);
 };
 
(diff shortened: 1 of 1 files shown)


stof:
anything missing here ?

aroc:
Hey @stof, sorry for the delayed response here! You're right, there's no need for this. Merging now. :)

PR #132, opened 2016-03-31
Removing deprecated pragma and renaming files to jsx

anthonymayer:


dustinlarimer:
@anthonymayer thanks for putting time into this- just curious about the reasoning behind the change. Is there a functional benefit to (re)naming files in this way?

anthonymayer:
@dustinlarimer: I don't know of any functional change of switching to `.jsx`, since build systems control the whole thing anyway. It's more just standard convention (from everything I've seen). I removed the pragma though because it's deprecated since React v0.12 (see https://facebook.github.io/react/blog/2014/10/16/react-v0.12-rc1.html). I ran into this, because I was trying to import components from keen-explorer into another project that uses babelify instead of reactify. I'm also guessing that a newer version of reactify would complain about the pragma, but I haven't tested it.

dustinlarimer:
@anthonymayer thanks for catching me up on those changes! I totally agree about removing the pragma- seems like a necessary change. I'll have to share the `.jsx` filename switch with my teammates though, as we're currently sketching out our internal style guide, linting, patterns, etc. We're also working on a hefty update and some new goodies that will be announced in the coming weeks, so I apologize in advance if this sits idle for awhile longer. Thanks again for your help!

anthonymayer:
So in case this PR is a little too much of an issue because of all the file renames, which I would totally get, I also put together https://github.com/keen/explorer/pull/134 quickly. If you're fine taking that one, feel free to drop this one.

aroc:
I just merged #134.  I think we can close this for now given that we don't have any short term intention of changing the file names to `.jsx`. Thanks for the work there @anthonymayer! We can revive this if necessary in the future.

PR #134, opened 2016-04-06
Removing deprecated pragma

anthonymayer:
This PR only removes the deprecated jsx pragma and some minor whitespace issues.


aroc:
Sorry we didn't get to this sooner! This is still worth doing, so we'll find a way to resolve the conflicts here. Thanks @anthonymayer!!

PR #136, opened 2016-04-07
Fixes 135

aroc:
Ensures the "optional" and "inverted" properties on the step objects are properly handled when being formatted after the API returns them.


sfalkoff:
🚢 Thanks for writing tests!

aroc:
:) thx for the review!

PR #140, opened 2016-04-25
Fixes #128: Funnels invalid date when choosing actor property

### LINKED ISSUE #128: Invalid time JS error in funnel analysis
When performing a funnel analysis it is not possible to select an actor property due to an Uncaught Error: Invalid time value (see screenshot).

keen-error

aroc:
## What does this PR do?

It fixes #128. The issue was that the global `timeframe` property was *not* being removed when migrating from another query type to a funnel query, however the `time` property *was* being removed. They *both* should have been removed.

The `timeframe` property is removed in the `_migrateTOFunnel` private function in the `ExplorerStore` now and a test has been added.

## How do I test this?

Read #128 to learn more.

1) Create a count query.
2) Run it.
3) Refresh the page.
4) Change to funnel query.
5) Try and choose actor property/finish creating and running the funnel query.


dustinlarimer:
:+1: looks solid to me, runs fine locally and not seeing any glaring syntax issues.

PR #141, opened 2016-04-26
Fixes #137

### LINKED ISSUE #137: I clicked the green button and got stuck
If you build a query and accidentally click the green button "save query" instead of "run query", you get an error about how you must name the saved query. This unfortunately is really easy to do because Save Query is the rightmost button and it's green. So instinctively I want to press it more than the blue "run query" on the left.

Unfortunately, now you are completely stuck because it won't let you click "run query" after that error appears.

You have to reload the explorer :(

Steps to repro:

1. Create a query like count last_14_days
2. Click save query instead of run query.
3. Get error.
4. Be stuck.

aroc:
## What does this PR do?
Fixes #137. Sets the "saving" property back to false after an attempted save when the model is invalid.

## How can I test it?

1) Fill out the query builder with a new query. Leave the "Query name" field above the visualization area blank.
2) Click "save". Observe that there is an error message saying that you must fill out the query name.
3) Click "run query".
4) Observe that the query successfully ran and the error message was cleared.


dustinlarimer:
👍 tidy

PR #144, opened 2016-04-26
Fixes #143: Better parsing of numeric filters

### LINKED ISSUE #143: Number filters parsed as strings on refresh
If you create a query with a filter value that's a `Number`, if you refresh, the value is parsed as a `String`.

aroc:
## What does this PR do?
It fixes #143. Specifically, when a filter `property_value` is checked for type on loading of the Explorer, we check to see if it's a `string` and if so, we then check to see if it's a `number` or not (using `isNaN`). If it's a `number`, we ensure the `property_value` of the filter is set to `Number` rather than `String`.

## How can I test this?
1) Create a query with a numeric filter.
2) Refresh the page.
3) Check the filter. Make sure it's still `Number`.


dustinlarimer:
👍 works as expected

PR #146, opened 2016-04-26
Fixes #145

### LINKED ISSUE #145: Global group_by not removed when switching to funnel analysis type
When you have a `group_by` property selected, then switch to a Funnel query, the global `group_by` property is not removed.

aroc:
## What does this PR do?
Fixes #145. Global group_by property not removed when switching to funnel queries. This ensures it's removed by adding the removal of that property to the private function `_removeInvalidFields` in the `ExplorerStore`.

## How can I test this?
1) Create a query with a `group_by` property. Run it.
2) Switch the analysis type to "funnel".
3) Notice that the global `group_by` property is no longer present in the URL.
4) Run the query. Inspect the request. Notice that the global `group_by` is not present.


dustinlarimer:
👍 works as expected

PR #147, opened 2016-04-26
Fixes #139: "Exists" filter breaks when changing the property name

### LINKED ISSUE #139: Error when you modify the property name of an "exists" filter
If you use the "property exists" filter and run the query, it works fine. However if you then go back to change the property to filter on, the filter seems to disappear or give an error.

Steps to Repro
1. Run a query like count "action" with a filter for ```<any property> exists true```. It works.
2. Now modify the filter property to any other property and run the query. You get an error ```The specified value is invalid for the given operator. Operator: exists. Value: true```

You can work around it by completely removing your filter and re-creating it.

Today, a Keen Pro customer also reported that it was impacting them.

screen shot 2016-04-20 at 10 58 06 am

aroc:
## What does this PR do?
It fixes #139. It ensures that if you change the property name on a filter with an `exists` oprerator, nothing else changes and you can still run the filter.

## How can I test this?
1) Create a query with a filter that has any property name, an exists operator, and either true or false for the value.
2) Run the query.
3) Change that filter's property name to something else
4) Run the query again.
5) Confirm it ran successfully.


dustinlarimer:
👍 works as expected

PR #148, opened 2016-04-27
Fixes #142

### LINKED ISSUE #142: Funnel list filters are not properly loaded from URL
If you build a funnel that has a step filtering using an "in" filter, then share the link with someone, the filter gets parsed as "[Object object]" and breaks.

Repro steps:
1) Create a funnel query. Make sure one step has a list filter. Something like "{properpty} matches any value in a list: [1, 2]
2) Refresh
3) Notice that the query didn't run. Click run.
4) BUG: Open the filters for the step with the list filter. Notice the property value is `[object Object]`.

aroc:
## What does this PR do?
This fixes #142 Funnel step filters not properly parsed from URL.

## How can I test this?
1) Create a funnel query and have one of the steps have a List filter.
2) Run the query. Notice it works as expected.
3) Refresh the page.
4) Open the filters panel for that filter step and notice the List filter is properly setup.


dustinlarimer:
👍 works as expected - nice work!

PR #152, opened 2016-06-01
Fix 3.0 handle schema timeout

dustinlarimer:
## What does this PR do?

This PR changes the way collection schema data is requested, fetching individual collection schema on demand, rather than fetching all schema for all collections on initial load from the project resource. This will reduce the initial load time and memory allocation for our largest projects, without introducing a noticeable lag for smaller projects.

## How should this be tested?

Load it up, confirm the tests work, run ad-hoc queries and saved queries, modify parameters, etc., and confirm there are no errors and the app works as expected.


client/js/app/actions/ExplorerActions.js
```diff
@@ -26,11 +27,17 @@ var ExplorerActions = {
   },
 
   update: function(id, updates) {
+    var q;
```
joannecheng: How do you feel about calling this variable `query`? Or updated_query?

PR #156, opened 2016-06-06
Fixes #155

### LINKED ISSUE #55: queries do not run when clicked on http://keen.github.io/explorer
The queries do not automatically run when they're clicked. 

http://keen.github.io/explorer/?query%5Bevent_collection%5D=page_load&query%5Banalysis_type%5D=count&query%5Btimezone%5D=UTC&query%5Btimeframe%5D=this_1_weeks&visualization%5Bchart_type%5D=metric#demo

### LINKED ISSUE #155: Saving query with "String" filter with integer string value loads the query with numeric property value
If you create a query with a `String` filter `contains` ("string contains") and make the value `500` then save the query, the query's filter gets saved with a numeric/integer property value rather than an String.

Reproduction:
1) Create a basic count query
2) Add a filter. "{property name} {contains} {1}". Make sure the property value number returns at least 1 result.
3) Save the query. Reload the page. Notice the returned result set is empty now. That's because if you dig deeper you'll see the property value of the filter is a _number_, 1, not the String 1. It's saved in Keen as a string, but Explorer typecasts it into a Number at load time.

aroc:
## What does this PR do?
It fixes #55.

## How can I test this?
1) Create a simple count query.
2) Add a filter with the `contains` operator and a String property value that's made of only numeric characters, like "1".
3) Run the query and ensure you get at least 1 result. If you don't - change the event collection or change the property value number to something else.
4) Save the query.
5) Reload the page. Run the query again. Ensure you get the same result.

## Does this have tests?
It sure does!
https://github.com/keen/explorer/compare/FIX-155?expand=1#diff-0f00529d4845b3fefd78d8fa03ecbc36R16


sfalkoff:
🚢

PR #160, opened 2016-06-30
Fix #125

### LINKED ISSUE #125: Geo filter "radius in miles" input doesn't work
image

aroc:
## What does this PR do?
It fixes #125.

## How can I test it?
Try creating a geo filter with all fields filled out. Ensure it runs as expected by checking both the request (ensure the filter is formatted correctly in the POST request) and the response for correct data with a sample project.


sfalkoff:
👍 Adding test coverage would also be great. Thanks for addressing this @aroc!

PR #166, opened 2016-07-23
Fix “recentEvents of undefined” errors coming from the EventBrowser

aroc:
## What does this PR do?
It fixes a bug where the `<EventBrowser>` component tries to access the recent events of a collection name that does not exist in the project schema.

## How can I test this?
I've added a couple tests to ensure this works as expected. So start by running the test suite. Then try using the EventBrowser a bunch. One thing to test is typing in a event collection name in the query builder area that does *not* match a real event collection for that project. Then open the Event Browser and make sure no errors are thrown.


sfalkoff:
🚢

PR #169, opened 2016-07-26
Ensure steps appear in correct order on funnel visualizations

sfalkoff:
## What does this PR do?

Resolves issue causing funnel steps to appear out of order in some visualizations. 

## How can I test this?

Get the Explorer running locally and try running several funnel analyses. Ensure that the order of the steps in your charts matches the order of the steps in your query.

### DIFF
--- a/client/js/app/components/explorer/visualization/keen_viz.js
+++ b/client/js/app/components/explorer/visualization/keen_viz.js
@@ -24,7 +24,7 @@ var KeenViz = React.createClass({
       .title(null)
       .type(this.props.model.metadata.visualization.chart_type);
 
-    if (!this.props.model.query.analysis_type === "funnel") {
+    if (this.props.model.query.analysis_type !== "funnel") {
       this.props.dataviz.sortGroups('desc')
     }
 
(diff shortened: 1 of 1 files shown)


dustinlarimer:
👍 perfect, nice work @sfalkoff

client/js/app/components/explorer/visualization/keen_viz.js
```diff
@@ -22,9 +22,12 @@ var KeenViz = React.createClass({
       .el(this.refs['keen-viz'])
       .height(300)
       .title(null)
-      .sortGroups('desc')
       .type(this.props.model.metadata.visualization.chart_type);
 
+    if (!this.props.model.query.analysis_type === "funnel") {
```
aroc: This should be:
```
if (this.props.model.query.analysis_type !== "funnel") {
```

In the current form, this line evaluates the truthyness of `this.props.model.query.analysis_type` and *then* compares that result (`true` or `false`) for strict equality against `"funnel"`. But what you want to do here is check for strict equality between two strings.

aroc:
👍

Issue #167, opened 2016-07-25
host parameter in embed functionality wrong and not required


spraetz:
If you click the embed for an ad-hoc query, you'll notice a "host" parameter in the setup.  That either needs to be removed or changed to include the version ("/3.0").

sfalkoff:
Resolved in https://github.com/keen/explorer/pull/171

PR #170, opened 2016-07-26
Remove optional host parameter from embeddable code sample

sfalkoff:
Resolves issue #167.

PR #171, opened 2016-07-29
Fix Issue #167

sfalkoff:
The goal of this PR is to remove the `host` property from the embeddable code sample without making any changes to the client. This is an update to https://github.com/keen/explorer/pull/170.


aroc:
👍

Issue #150, opened 2016-05-23
Embed code for percentile query does not include "percentile" property value 


subbuthepeaceful:
When you click on the "Embed" button to embed the results of a Percentile query, the resulting Embed code does not contain the percentile value used in the query.

Originally reported by John Eskilsson, replicated along with screenshot.

<img width="973" alt="percentilequeryembed" src="https://cloud.githubusercontent.com/assets/460078/15478949/8317b4da-20d2-11e6-845a-767ea75db844.png">

PR #173, opened 2016-08-05
Fix issue #150

sfalkoff:
Resolves https://github.com/keen/explorer/issues/150. The embeddable code sample for percentile queries now includes the percentile property and value. 

**Note:** The percentile value is currently being saved and executed as a string (https://github.com/keen/explorer/issues/168) instead of a number. Planning to fix that issue and roll the changes out together to prevent additional confusion for folks copy/pasting the code sample.


aroc:
👍

PR #174, opened 2016-08-06
Fix issue #168

sfalkoff:
### What does this PR do?

Resolves issue https://github.com/keen/explorer/issues/168 ensuring the percentile parameter is saved and executed as a number (not a string). Goes with bug fix: https://github.com/keen/explorer/pull/173.

### How can I test this?

Run a percentile query and check out the Request Payload in your browser (Network tab in Chrome). The percentile value should be a number.


client/js/app/stores/ExplorerStore.js
```diff
@@ -283,6 +283,10 @@ function _wrapGroupBy(group_by) {
   return group_by;
 }
 
+function _coerceToNumber(str) {
+  return str && typeof(str) === 'string' ? Number(str) : str;
```
aroc: You can just return `Number(str)` here, as `Number({a string value})` is the same as `Number({a number value})`.

sfalkoff: Thanks for reviewing @aroc. I included the conditional to account for a situation where an empty string is passed to `_coerceToNumber`. We're currently using an empty string as the default percentile value, which will cause the query to error if executed. Without the conditional, `Number("")` returns `0` which will successfully return unexpected query results to the user (when there should probably be an error).

It looks like there is another issue with validation right now, making it possible for a user to execute a percentile query without entering a percentile value in the input field, actualizing this edge case. I'm happy to look into that too (perhaps as a separate PR). Thoughts?

aroc: Yeah I'd take a look at that in a separate PR.

Good catch on the empty string. Is the empty string *the only* other value this function will be called with? If so, then I'd be more explicit about it and do a check for an empty string instead. Ex: `return str === "" ? str : Number(str)`. And then you likely remove the need for the `_coerceToNumber` entirely, as it's only wrapping a very small piece of logic used in only 2 spots. You can also do `parseFloat(str) || str` as `parseFloat` will return `NaN` for empty strings.

sfalkoff: There might be a situation where we're calling this function with `null` as well -- I will take another look to confirm.

sfalkoff: Yup, looks like we need to account for both `null` and empty strings. I'd like to move forward by removing `_coerceToNumber` and using `parseFloat()` directly on lines 306 & 320: 

306: `newAttrs.query.percentile = parseFloat(newAttrs.query.percentile);`
320: `newModel.query.percentile = parseFloat(newModel.query.percentile);`

If the percentile is `NaN`, the query will error and the chart will continue to hang in a loading state (consistent with current behavior).

aroc: ok great. I think that makes sense. We can add percentile validation as another piece of work. :)

aroc:
👍

Issue #138, opened 2016-04-20
Keyboard controls for creating filters


wetzler:
A request came in from a frequent user of the explorer:

When setting filters or group by in the modal, make it so when you hit enter from a text field, it closes the modal. thx!

adamkasprowicz:
This one was pretty complicated from the UI/React.js side.
Good news - this feature is available in keen-explorer@5 (#239)
It will be deployed on the keen.io website in the next 1-2 weeks

PR #175, opened 2016-08-08
Feature request #138

0b3r:

Added feature requested in #138.  This allows the user while filtering to hit enter from a text field, apply the filter, and close the modal.
## How should this be tested?

I checked the test coverage for the modal and there is already a close test case.  This addition only utilizes functionality already covered.  There is no parameters passed so test case does not need to be expanded.

Currently this feature is only applied to text inputs as specified in the ticket.
## Related tickets?
#138


aroc:
Great, thanks for this @0b3r! We'll take a look as soon as we can.

PR #177, opened 2016-08-25
Add percentile value validation

sfalkoff:
### What does this PR do?

It's currently possible for a user to run a percentile query without a percentile value (a required field). This results in an error and the query appearing to load indefinitely. This PR addresses that issue by adding validation to notify users when the percentile value is missing.

### How can I test this?

- Run a percentile query with no percentile value and confirm that an error is displayed in the UI. 
- Run a percentile query with a percentile value and confirm that it executes as expected.
- Confirm all tests pass.

### DIFF
--- a/client/js/app/validations/ExplorerValidations.js
+++ b/client/js/app/validations/ExplorerValidations.js
@@ -55,7 +55,7 @@ module.exports = {
     },
     
     validate: function(model) {
-      return typeof model.query.percentile ==='number';
+      return typeof model.query.percentile === 'number';
     }
 
   },
--- a/client/js/app/validations/ExplorerValidations.js
+++ b/client/js/app/validations/ExplorerValidations.js
@@ -46,6 +46,20 @@ module.exports = {
 
   },
 
+  percentile_value: {
+    
+    msg: 'Choose a Percentile Value.',
+
+    shouldRun: function(model) {
+      return model.query.analysis_type === 'percentile';
+    },
+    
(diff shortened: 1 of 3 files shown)


client/js/app/validations/ExplorerValidations.js
```diff
@@ -46,6 +46,20 @@ module.exports = {
...
+  percentile_value: {
+    
+    msg: 'Choose a Percentile Value.',
+
+    shouldRun: function(model) {
+      return model.query.analysis_type === 'percentile';
+    },
+    
+    validate: function(model) {
+      return typeof model.query.percentile === 'number';
```
aroc: Given the name of the error message, I think the validation check here should be if it exists or not, rather than whether it's a `Number` type. Thoughts?

sfalkoff: Checking for existence sounds like a good thing to do here, thanks for the suggestion. Since we currently use `null` (set in the store) and `''` (in the common input component) as default values, I'd advocate for an explicit return statement that also checks whether the property value is a `Number` type. What do you think of this approach?

`return model.query.percentile !== null && model.query.percentile !== '' && typeof model.query.percentile === 'number';`

aroc: Yep I think that works well! 👍

sfalkoff: Great! Updated. Merging.

PR #178, opened 2016-08-26
FIX: Display an error message when a user marks a funnel's first step as optional

joannecheng:
Catch error in `ExplorerUtils#runQuery` and display notification when we get an error back from keen service


client/js/app/utils/ExplorerUtils.js
```diff
@@ -176,6 +176,9 @@ module.exports = {
           config.success(res);
         }
         if (config.complete) config.complete(err, res);
+      })
+      .catch(function(err) {
```
aroc: This can be written as:
```
.catch(config.error);
```

joannecheng:
@aroc rewrote code according to comment, could you re-review?

aroc:
looks good 👍

Issue #151, opened 2016-06-01
[Feature request] Add ability to clone a saved query


stof:
It would be helpful to be able to clone a saved query. The idea would be to populate the explorer with the settings of the saved query, and letting me save it as a different saved query (after doing some changes) instead of updating the existing query.

This is helpful when I have several similar queries to create (just changing one of the filter condition in this case, while keeping all other filters and other things the same)

aroc:
This would be a great feature and we've had it requested several times. Definitely something on our radar! Thanks for sharing your feedback with us :D

sfalkoff:
@stof this feature is live! You should be able to clone saved queries now. 

Thanks again, @menismu, for your contribution implementing this feature in #180.

PR #180, opened 2016-09-06
Implemented the clone function:

menismu:

It implements a Clone function for saved queries, the user is able to clone a saved query to later apply some changes and save the cloned query as a new query.
## How should this be tested?

Thinking in UAT this should be tested, first creating or saving a query and after that use the Clone button to make a clone of the saved query and check the cloned query can be modified before create a new one.

Also, it is important to mention tests have been created following the same guidelines I´ve seen in the code, most important files modified are ExplorerActions.js and ExplorerStore.js from my point of view but also query_actions.js and index.js in components/explorer path.

In terms of Development, I´ve tried to be as less intrusive as possible and also I made some assumptions like only the query attributes of a saved query is cloned, nothing related to id, name or other attributes not in query attribute of the explorer object. A function _cloneAttrs has been created in ExplorerStore.js just in case is needed to copy more attributes, it will be easier to only modified this funciton which should return the attributes to be merged by the _create function. Also, it could be important to mention to clone the query, first the main function in ExplorerStore.js is getting the data of the source explorer object to later copy the relevant attributes and finally call to _create function passing the copied attributes to create a new explorer object with the relevant attributes. It is also important to say that the dispatcher is used to populate a clone event has been raised from the Clone button, following the same implemented approach than create a new explorer object or query.

There are a couple of things I´m not 100% sure, one is the message shown to the end-user when Clone button is used, and the other one is the attributes copied, not sure if other attributes should be copied or only with the query object inside of a explorer object is enough.
## Related tickets?

I´ve made the decision to implement this feature because I´ve seen the ticket #151 where it was requested and I´m expecting also to use this function soon in one of the project I´m involved and I will use Keen platform to collect data.

### DIFF
--- a/client/js/app/stores/ExplorerStore.js
+++ b/client/js/app/stores/ExplorerStore.js
@@ -100,6 +100,17 @@ function _defaultStep() {
   }
 }
 
+/**
+ * Clone the query attributes to a new object.
+ * @param {Object} source  Source object from get the data to be copied.
+ * @returns {Object} Copy of the query attributes of th source object
+ */
+function _cloneAttrs(source) {
+	return {
+	    query: source.query
+	};
+}
+
 function _validate(id) {
   RunValidations(ExplorerValidations, _explorers[id]);
 }
@@ -534,6 +545,12 @@ ExplorerStore.dispatchToken = AppDispatcher.register(function(action) {
       action.models.forEach(function(model) {
         _explorers[model.id] ? _update(model.id, model) : _create(model);        
       });
+      finishAction();
+      break;
+      
+    case ExplorerConstants.EXPLORER_CLONE:
+      var source = ExplorerStore.get(action.id);
+      _create(_cloneAttrs(source));
       finishAction();
       break;
 
--- a/client/js/app/actions/ExplorerActions.js
+++ b/client/js/app/actions/ExplorerActions.js
@@ -26,6 +26,18 @@ var ExplorerActions = {
       models: models
     });
   },
+  
+  clone: function(sourceId) {
+	AppDispatcher.dispatch({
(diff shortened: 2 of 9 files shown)


aroc:
Just attaching some screenshots for others to see:

Clone button showing in button tray (only shows *after* a saved query has been selected):
image

Successful clone. Notice clone button is now gone as this is no longer a saved query. The query is set to active and the "build" panel is properly activated. A message is also shown at the alert area.
image

client/js/app/stores/ExplorerStore.js
```diff
@@ -100,6 +100,17 @@ function _defaultStep() {
...
 }
 
+/**
+ * Clone the query attributes to a new object.
+ * @param {Object} source  Source object from get the data to be copied.
+ * @returns {Object} Copy of the query attributes of th source object
+ */
+function _cloneAttrs(source) {
+	return {
+	    query: source.query
```
aroc: Quick note here: Objects are passed by reference so this actually wouldn't properly clone the query. You need to use something like lodash's `_.cloneDeep`.

sfalkoff:
@menismu thanks a lot! I branched off of this branch and am going to make a few updates over [here](https://github.com/keen/explorer/tree/clone-saved-query-feature). Feel free to pitch in if you like, but we're also happy to run with it now too - thanks again! 😄

client/js/app/stores/ExplorerStore.js
menismu: Thanks for letting me know @aroc.

sfalkoff:
@menismu I made a few modifications and deployed this from #193. Thank you again for your contributions - this feature is now live in production! 🎉

Issue #154, opened 2016-06-02
Filter property name with [ square bracket throws error


dustinlarimer:
Filter property name with [ square bracket throws error:
> Uncaught SyntaxError: Invalid regular expression: /request.params.filter[/: Unterminated character class

<img width="743" alt="screen shot 2016-06-01 at 9 36 23 pm" src="https://cloud.githubusercontent.com/assets/180438/15734117/f82f17a2-2840-11e6-8f6c-900532320f92.png">

menismu:
I´m happy to help fixing this bug if it is not yet in the development plan but I would need some indications to see how this component should work, should it accept characters like [ or ]? Is there any other character not allowed?

aroc:
@menismu Yes, we want to allow `[` and `]` characters. The full rules for what's allowed in a Keen property name are here in our docs: https://keen.io/docs/api/#property-name-rules

jacklittleton:
@menismu Keen IO internal resources are beginning to look at this; have you made progress on this, or should we continue? We wouldn't want to have a duplicate effort situation.

menismu:
@jacklittleton I got the fix done but I wanted to also write tests, if you and Keen don't mind to wait few days I will send a pull request before the end of the week with this bug fixed as well as regression test.

jacklittleton:
@menismu Hey, thanks for the PR review! Greatly appreciated!

PR #181, opened 2016-09-22
Issue 154 reg exp invalid in event name

jacklittleton:
## What does this PR do?

It escapes the parameter sent to `String.search()` in `ReactSelect.getItems()` so that RegExp reserved characters are properly escaped prior to searching.

## Why?

Without escaping, reserved RegExp characters like `[`, `{`, and `(` in a select element's search box would cause a SyntaxError.

## Tickets

https://github.com/keen/explorer/issues/154


test/unit/components/common/react_select_spec.js
```diff
@@ -49,5 +49,47 @@ describe('components/common/react_select', function() {
...
+      TestUtils.Simulate.focus(this.component.refs.input);
+      TestUtils.Simulate.change(this.component.refs.input);
+
+      var listItems = _.map(ReactDOM.findDOMNode(this.component.refs.list).childNodes, function(item) {
+        return item.textContent;
+      });
+      assert.sameMembers(listItems, ['THREE{tres}']);
+    });
+
+    it('should display items with curly braces (RegExp reserved char)', function() {
```
sfalkoff: Thanks for adding test coverage. Looks like "curly braces" should be "parentheses" on line 81.

tbarn:
Hey @menismu! I know you had been looking into this issue. We want to merge it soon, could you help us review it?

test/unit/components/common/react_select_spec.js
```diff
@@ -49,5 +49,47 @@ describe('components/common/react_select', function() {
       });
       assert.sameMembers(listItems, ['ONE']);
     });
+
+    it('should display items with square bracket (RegExp reserved char)', function() {
```
menismu: I'm wondering if the test cases should be increased to test also ] and ), it is probably a not real situation but as ReactSelect is a common or general component it will increase confident for future changes.

test/unit/components/common/react_select_spec.js
aroc: +1

test/unit/components/common/react_select_spec.js
aroc: +1. Adding a test for `)` might be nice here for better coverage.

PR #185, opened 2016-10-19
Use webpack instead of gulp

joannecheng:
### What does this PR do?

This PR replaces some of our gulp build steps with webpack and adds npm scripts for our common gulp tasks. We're switching to webpack to standardize and simplify our build process for our JS projects.

This PR also updates pickadate (because the version we used was incompatible with webpack), and fixes some accidental global variables in our codebase.

### How to test:
* Run `npm install`. There are quite a few new modules.
* Run
  * `npm run dev`
  * `npm run build`
  * `npm run minify`
  * `npm run production`
and make sure they work correctly.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,19 +1,24 @@
-'use strict';
-
 const webpack = require('webpack');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+
+const buildName = 'keen-explorer';
 
 const devModule = {
-  entry: './client/js/app/app.js',
+  entry: ['./client/js/app/app.js', './client/styles/base.less'],
+
   output: {
-    filename: 'keen-explorer-webpack.js',
-    path: './dist/'
+    filename: `${buildName}.js`,
+    path: './dist/',
+    chunkFilename: `${buildName}.js`
   },
+
   module: {
     loaders: [
       {
         test: /picker/,
         loader: 'imports?define=>false'
       },
+
       {
         test: /\.js$/,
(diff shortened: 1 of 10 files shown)


jacklittleton:
**NOTE** (for the webpack impaired, like me): Do the following to test this:

1. Run `npm install -g webpack` (may not be necessary if you've previously done this)
2. Just run `webpack` from the explorer root directory (where the `webpack.config.js` file resides)

webpack.config.js
```diff
@@ -0,0 +1,46 @@
+const webpack = require('webpack');
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+
+const buildName = 'keen-explorer';
+
+const devModule = {
+  entry: ['./client/js/app/app.js', './client/styles/base.less'],
+
+  output: {
+    filename: `${buildName}.js`,
```
aroc: When we do a webpack production build we'll want to add a `.min` to this filename. The easiest way I've found to do this is to just check the environment and manually add a `.min` when building for production. Example from statics: https://github.com/keenlabs/keen-statics/blob/master/webpack.config.js#L9

jacklittleton:
General observation: If I `npm run dev`, and then go to http://localhost:8081, I see a directory listing of the root source directory, not the explorer. The README says:

> You can now view the demo locally at http://localhost:8081/ or your specified port.

joannecheng:
@jacklittleton I added the server gulp task back so we can keep the same URL. Thanks for catching!

PR #187, opened 2016-10-21
Collection name bad chars, no props

jacklittleton:
This fixes a bug that would occur when an event collection name had certain characters (namely /, but also potentially #, ?, :, & and <space>). If that condition was met, the Target Properties drop down would display empty, even though there were properties associated with that event collection.

## How should this be tested?
Additional unit tests were created to cover the /, #, ?, :, &, and <space> conditions. To test in the browser, create event collections with each of the offending characters, and check that their properties show in the Target Properties drop down after the collection is selected.

- [X] Does the code compile without warnings (check shell, console)?
- [X] Do all tests pass?
- [X] Does the UI, pixel by pixel, look exactly as expected (check various screen sizes, including mobile)?
- [X] If the feature makes requests from the browser, inspect them in the Web Inspector. Do they look as expected (parameters, headers, etc)?
- N/A If the feature sends data to Keen, is the data visible in the project if you run an extraction (include link to collection/query)?
- N/A If the feature saves data to a database, can you confirm the data is indeed created in the database?


sfalkoff:
🚢

Issue #121, opened 2016-02-02
"Interval" doesn't clear properly and makes it so you can't save funnel queries.


wetzler:
At least 2-3 times today I've had a interval of "minutely" mysteriously appear in the explorer. One time it was when updating a saved query, it changed from monthly to minutely.

Another time, it was after I clicked "new query" that the new query started with a default interval of "minutely".

Another time, it appeared after I clicked "clear". 

Sorry I don't have more details yet!

wetzler:
Ok I'm pretty sure if you CLEAR any query with an interval, it reset it's interval to "minutely".

dustinlarimer:
Adding a bit more info here: the Interval component will always have a value, but should be toggled "off" when the query builder is cleared out.

wetzler:
This came up again when building a demo for a customer. I created a funnel and then tried to save it. I got an error about "interval" being set.

Steps to repro:
1. Create project and load sample table
2. Run a count of login and signup events, previous 7 days, daily
3. Build a funnel with step 1 signup, steps 2-5 login
4. Click save funnel. 
5. Get an error about interval being set. 

![screen shot 2016-06-07 at 3 31 13 pm](https://cloud.githubusercontent.com/assets/1840448/15876921/51c7f26a-2cc5-11e6-918b-a6c5150c66a0.png)

sfalkoff:
Shipped a fix for the issue saving funnel queries in #188. Toggling the Interval component "closed" on "Clear" is still outstanding, but should not impact users' ability to save queries.

sfalkoff:
Opened a new issue for the interval component toggle behavior here: #189.

PR #188, opened 2016-10-21
Remove invalid interval parameter from funnel queries

sfalkoff:

The `interval` parameter is not supported by [funnels](https://keen.io/docs/api/#funnels). This PR removes the invalid `interval` parameter from the Explorer model for this type of query. 

This change resolves the issue saving funnel queries reported in #121. The changes in this PR do not address the issue of toggling the Interval component to a "closed" state when the query builder is cleared out (also reported in #121); I recommend addressing those in a separate PR.

## How should this be tested?

I added a unit test for this change. To test these changes in the browser, create a funnel query and confirm that there is no `interval` set in the URL.


joannecheng:
👍

Issue #94, opened 2016-01-13
Feature request: export to CSV from Table view


nemo:
Hey folks,

I think it's an obvious extension to the table view of the data to have an `Download as CSV` button of some sort.

This will make it easy to do further analysis in Excel/Spreadsheets outside of Keen (for example, when you want to perform rate of change and/or accumulation).

wetzler:
Hey! Great feature request. In the meantime, I wanted to let you know about this google spreadheets hack that always you to get your data into tables relatively quickly:
https://www.linkedin.com/pulse/simplest-way-import-your-keen-io-data-google-sheets-gabellieri

wetzler:
One more thing, it's a little bit awkward to copy & paste tables out of Keen, but it is possible. Try starting from the bottom right corner and scrolling up (but not too far) to select the table contents and then copy & page into excel.

nemo:
Thanks @wetzler – ended up figuring out how to copy it from the table. Still would be great to have the button :P

dustinlarimer:
@nemo @wetzler this is now in production- thanks for the nudge! Enjoy!

PR #190, opened 2016-12-03
- Added a button to export to CSV

menismu:
What does this PR do? How does it affect users?

It implements an Export to CSV function for for table chart type, the user is able to use the new Export button to open or download a CSV file.

How should this be tested?

Thinking in UAT this should be tested, first making sure the button only appears when Table chart type is selected for the current query loaded.

I´ve created some tests checking ExportToCsv function is called as well as ensuring Export button only appears when table chart type is selected.

There few things to be taken into account, I´ve used a very simple way to generate the CSV file, this could be improved integrating a library if really needed, this code is in a new file created DataUtils.js and this is the only thing not tested in the spec files in the solution. The other bit I´m not quite sure it is the best solution is related to the check if the dataviz shown in the table type, it is checked through the metadata in the model (model.metadata.visualization.chart_type) but it could be (probably) done by dataviz.type.

Related tickets?

I´ve made the decision to implement this feature because I've seen the ticket #94 where it was requested.

### DIFF
--- a/client/js/app/components/explorer/visualization/chart.js
+++ b/client/js/app/components/explorer/visualization/chart.js
@@ -66,7 +66,8 @@ var Chart = React.createClass({
 		}
 		else {
 	  	chartContent = (
-	  		<KeenViz model={this.props.model} dataviz={this.props.dataviz} />
+	  		<KeenViz model={this.props.model} dataviz={this.props.dataviz}
+	  				exportToCsv={this.props.exportToCsv}/>
 	  	);
 	  }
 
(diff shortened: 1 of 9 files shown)


menismu:
I´m more than happy to listen feedback about the changes, as I've mentioned in the pull request first comment, I've not sure about a couple of things and if changes needed just let me know.

dustinlarimer:
@menismu thank you so much for your amazing contributions! I'm sincerely sorry for the delay in reviewing your PRs. We're catching up now and will get this reviewed :+1:

dustinlarimer:
@menismu I've made a few modifications and will be deploying this from #194 shortly. Thanks again for your hard work on this! Super excited to get this released!

dustinlarimer:
@menismu this is now live in production - thanks again for your help in making this happen!

PR #195, opened 2016-12-21
Use snake_case property names in embeddable code sample

sfalkoff:
# What does this PR do? 

This PR removes camelCase conversion for properties provided to the `Keen.Query` object in the embeddable code sample. For example, `eventCollection` now displays as `event_collection`, the format that the [API requires](https://keen.io/docs/api/). 

This change is compatible with both [keen-js](https://github.com/keen/keen-js) and [keen-analysis.js](https://github.com/keen/keen-analysis.js) (whereas camelCase is only compatible with keen-js, which converts camelCase parameters to snake_case/underscore format before being sent to the API).

## Before

before

## After

after

# How should this be tested?

Get this branch running locally, and try running several queries in your browser. Open the embeddable code sample for each query, and confirm that the query properties passed to `Keen.Query` are in snake_case format where applicable (e.g. `event_collection`, `group_by`, `target_property`).


sfalkoff:
Thanks @dustinlarimer!

Issue #192, opened 2016-12-20
Select field displays incorrect visualization type after saving query


sfalkoff:
Steps to reproduce:

1. Run a new count query. Notice that the visualization type is set to "Metric" by default.

![screen shot 2016-12-20 at 11 11 31 am](https://cloud.githubusercontent.com/assets/6183404/21364408/99da4ef2-c6a5-11e6-9df3-9dc61aa98dab.png)

2. Enter a name for your query and save it. Notice that the visualization still displays with the "Metric" chart type (expected), however, the select component shows "Bar" and "Metric" no longer appears in the drop down (not expected). 

![screen shot 2016-12-20 at 11 11 55 am](https://cloud.githubusercontent.com/assets/6183404/21364499/f8371548-c6a5-11e6-9f3d-0943be1d5e8e.png)

I've also seen this issue after saving count_unique queries -- it may be happening with other query analysis types as well.

josephwegner:
Picking this up

dustinlarimer:
@josephwegner @sfalkoff is this resolved?

dustinlarimer:
This appears to be working correctly in prod now- closing 👍

PR #196, opened 2016-12-22
remove null values when we turn the group_by into an array

josephwegner:
Saved queries will represent `group_by` as a `null` for queries without a group_by value. Explorer tries to be smart by turning non-array `group_by` values into single-element arrays, but makes the poor mistake of wrapping `null` `group_by`s into an array. That messes things up down the line, specifically causing Explorer to think the query is categorial (has a group_by).

This change removes `null` values from the `group_by` array. Effectively, for users, that means they will get the correct chart type options for their `singular` saved queries.

## How should this be tested?

Pull it down, and create a regular ol' count query with a metric type. Save it. The chart type should still be metric. If it's not we have trouble.

I'd also suggest trying a few other query types, just to watch for unintended consequences.

- [ ] Do all tests pass?

## Related tickets?
#192 (JIRA #520)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "classnames": "2.1.2",
     "flux": "^2.0.3",
     "json-stable-stringify": "1.0.0",
-    "keen-analysis": "^1.1.0",
+    "keen-analysis": "1.1.0",
     "keen-dataviz": "^1.0.4",
     "keymirror": "~0.1.0",
     "lodash": "4.0.0",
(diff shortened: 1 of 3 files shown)


josephwegner:
Oh, I forgot to mention, I pinned our `keen-analysis` to 1.1.0... Dustin has a bug that he's working on in the recently released 1.2.0 that was causing test failures.

package.json
```diff
@@ -8,7 +8,7 @@
     "classnames": "2.1.2",
     "flux": "^2.0.3",
     "json-stable-stringify": "1.0.0",
-    "keen-analysis": "^1.1.0",
+    "keen-analysis": "1.1.0",
```
sfalkoff: Thanks for the heads up about the keen-analysis version. This just in: looks like it's being addressed in https://github.com/keen/explorer/pull/197.

josephwegner: @sfalkoff His PR should overwrite my pinning, so I'm gonna leave this in (that will avoid CI/build errors depending on which merges first)

PR #200, opened 2017-01-06
Revert to previous version of PickADate

aroc:
# What does this PR do?
It reverts the vendor files for the pick-a-date datepicker UI library back to `3.5.4`. `3.5.6` which we updated to does not work with Explorer for reasons not yet known.


dustinlarimer:
Deployed latest built files to our CDN and will update the README in my last commit here.

Issue #203, opened 2017-03-16
Missing timezone


SegFaultx64:
Hi folks,

The explorer is missing the Atlantic Timezone, which is not a huge problem normally since few people live in it. However right now (during DST) -4GMT becomes the time zone in the eastern United States. This means that we basically can't correctly run queries right not :sob:.

dustinlarimer:
@SegFaultx64 thank you for taking time to put #205 together! These timezone records correspond to the list of named timezones supported by the API: https://keen.io/docs/api/#named-timezones, so I'm afraid the update won't have the desired effect. 

I just merged #206 which will set the local timezone appropriately, either selecting a timezone with the corresponding offset, or setting the explicit numerical offset as a backup. I think this should resolve the issue for you. Please let me know if that's not the case!

Issue #153, opened 2016-06-01
Unclear if timezones respect DST


josephwegner:
To be clear, the timezone picker _does not_ respect DST. That may be the actual issue here, but it's up for debate.

Anyways, when you have an actual timezone chosen (ie: "Europe/Stockholm") the TZ picker gives no indication of whether or not it is respecting DST. In that case, it is my assumption that I should think like a human, and expect it to run the query in DST. For Europe/Stockholm, that is +2.

<img width="471" alt="screen shot 2016-06-01 at 4 04 59 pm" src="https://cloud.githubusercontent.com/assets/657707/15725761/fd339c74-2812-11e6-87dd-370377b9494e.png">

HOWEVER, when you're actually choosing your timezone, the dropdown options do specify the GMT offset. In that case it's a bit more clear that DST won't be respected (assuming you know what the offset _should_ be).

<img width="474" alt="screen shot 2016-06-01 at 4 06 27 pm" src="https://cloud.githubusercontent.com/assets/657707/15725768/02e8c9fa-2813-11e6-90f2-3d23149f1b81.png">

It probably makes the most sense just to include (GMT+1) thing in the selected state - then it's always clear.

BlackVegetable:
DST isn't respected as you can select an illegal hour via the explorer (the one we skip over) and the hour that is duplicated isn't viewable.

dustinlarimer:
@josephwegner @BlackVegetable the timezones in this component correspond to our [named timezones](https://keen.io/docs/api/#named-timezones), which should be adjusted for DST. #206 should also help here.

BlackVegetable:
@dustinlarimer Technically they aren't flawless. I can select 2:30 AM on Sunday, March 12th, 2017 US/Mountain time. That hour never occurred because it was skipped over by DST. Similarly, I can select Sunday, November 6th, 2016 at 1:30 AM where it is unclear *which* 1:30 AM I intended and I can only select one of these two from the dropdown (because the hour is repeated...)

PR #204, opened 2017-03-16
Add Atlantic time

SegFaultx64:
# What does this PR do?

Add a new timezone to the drop down

I selected Caracas as it appears to be the largest city in that timezone. This is critical to allow users to make queries of the eastern US during DST.

# How should it be tested?

IDK since it is a one line change to a json object. Probably check that I spelled Caracas right...

# Related issues

#203, #153


SegFaultx64:
Fixes #203

PR #205, opened 2017-03-17
Add Atlantic time

SegFaultx64:
What does this PR do?

Add a new timezone to the drop down

I selected Caracas as it appears to be the largest city in that timezone. This is critical to allow users to make queries of the eastern US during DST.

How should it be tested?

IDK since it is a one line change to a json object. Probably check that I spelled Caracas right...

Related issues

#203, #153


dustinlarimer:
Cross-posting from #203 for reference: 

> These timezone records correspond to the list of named timezones supported by the API: https://keen.io/docs/api/#named-timezones, so I'm afraid the update won't have the desired effect. 

> I just merged #206 which will set the local timezone appropriately, either selecting a timezone with the corresponding offset, or setting the explicit numerical offset as a backup. I think this should resolve the issue for you. Please let me know if that's not the case!

Thanks again for taking the time to work on this- much appreciated! Please let me know if the update in #205 doesn't help. I'll be deploying this shortly.

PR #206, opened 2017-03-17
Find and set local timezone by default

dustinlarimer:

This PR adds logic to find the first timezone value matching the user's local timezone offset. This also corrects for DST, shifting the offset ahead an hour when necessary to ensure the proper match.

screen shot 2017-03-17 at 4 28 20 pm

## How should this be tested?

Load it up and take it for a spin.

## Related tickets?

This will resolve a persistent issue where charts appear to be "wrong" for users in the US. This happens when timestamps are localized, making days and months display one whole unit earlier than what is expected. Eg: March 1st (midnight) displays as February 28th (-7hr). By setting the proper timezone, the results are shifted in the other direction.

### DIFF
--- a/client/js/app/stores/ExplorerStore.js
+++ b/client/js/app/stores/ExplorerStore.js
@@ -39,7 +39,7 @@ function _defaultAttrs() {
       percentile: null,
       group_by: [],
       interval: null,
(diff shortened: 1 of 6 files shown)


aroc:
Reviewing...

Issue #59, opened 2015-11-19
Filters on saved queries with property_value strings can be interpreted as dates


dustinlarimer:
If a filter's `property_value` can be parsed into a valid `Date` object, that filter value will get handled as such. Eg: "2015" becomes "2015-11-19T00:00:00". This occurs prior to saving a given query, meaning the saved query parameters are also incorrect.
1. Create a filter for a property where `operator` equals a string value of 2015
2. Close filter modal
3. Name and save the query
4. Reload the query and re-open the filter modal
5. That filter's property value is now "2015-11-19T00:00:00".

<img width="432" alt="screen shot 2015-11-19 at 12 55 23 pm" src="https://cloud.githubusercontent.com/assets/180438/11284090/d561f53c-8ebc-11e5-85bd-b42efd5b0e57.png">

<img width="436" alt="screen shot 2015-11-19 at 12 55 15 pm" src="https://cloud.githubusercontent.com/assets/180438/11284096/ddf5c30e-8ebc-11e5-943a-1bf614478109.png">

aroc:
As part of this fix, we should make the change to only allow `Datetime` coercion type on `keen.timestamp` and `keen.created_at`, as datetime's only work on those properties anyway.

PR #209, opened 2017-03-21
Prioritize boolean string values when parsing filters from URL params

aroc:
## What does this PR do?
It adds another check in the `getCoercionType` function to prioritize `"true"` and `"false"` strings as `Boolean` filters.

## Why?
The chances that a user wants a filter to be a `Boolean` when the `property_value` is a string that is exactly `"true"` or `"false"` is higher than the chances that they want it to remain a string.


dustinlarimer:
Is this issue also stemming from this method? https://github.com/keen/explorer/issues/59

aroc:
@dustinlarimer nope, because we don't parse Datetime's until lower down in the code here. Same function, but unrelated.

dustinlarimer:
Right- figured it was unrelated, but perhaps possible to hit 2 in 1 shot? :)

aroc:
Certainly possible! But I'd say let's get this in first - we have a customer impacted right now that we can make very happy.

dustinlarimer:
👍🏻

aroc:
Just took a look at #59 - given how we do the "is this a datetime" check, I think the best solution is the URL "type" param we spoke about. Plus it would solve this whole type thing once and for all! 🎉

PR #210, opened 2017-03-21
Fix props assignment bug with multi group by

aroc:
## What does this PR do?
It fixes a bug when trying to toggle on the multi-group by query feature. We don't want to assign the value to `this.props.value` here, as it will throw an error in strict mode.


aroc:
No longe relevant. This was fixed by some other work.

PR #211, opened 2017-04-03
fix typo in timezone mouseover

akahncmu:
because spelling


akahncmu:
i really dont need all 4 of you to look at this, whomever gets there first.

sfalkoff:
@akahncmu thanks for adding the minified files. `keen-explorer.min.js.gz` is an unnecessary file that shouldn't be generated with the build anymore (looks like this branch may be behind master); if you could remove that file and resolve merge conflicts, these changes will be ready to go.

PR #213, opened 2017-05-12
Allow user to select what fields to see in an extraction table

joannecheng:
## What does this PR do?

This PR lets users filter what properties are displayed in the explorer during an `extraction` query.

For example, if I have a large number of extraction properties, and I only want to view `email` and `user id` properties, I would use this feature to only display the two properties I want to see.

## Why?

Sometimes users only need to see one or two columns out of many in the explorer table view.

## Docs
https://docs.google.com/document/d/1OMae4kp1wu9Sn4owT8hYkCL2-95gyc_qJulIEuccdXk/edit#

## Demos

filterextractionoptions


client/js/app/components/explorer/query_builder/extraction_properties_filter.js
```diff
@@ -0,0 +1,45 @@
+var React = require('react');
+var ReactMultiSelect = require('../../common/react_multi_select.js');
+var ExplorerActions = require('../../../actions/ExplorerActions');
+
+var ExtractionPropertiesFilter = React.createClass({
+
+  _getKeys: function() {
+    var keys = _.keys(this.props.result);
+    var keyList = _.map(keys, function(key) {
```
aroc: Where is `_.keys`, `_.map` etc coming from in this file? I don't see `lodash` required in the file. Also, we probably shouldn't use `lodash`. We have `es2015` babel in this project so we can use `Object.keys`, `map`, `filter`, etc.

client/js/app/components/explorer/query_builder/extraction_properties_filter.js
```diff
@@ -0,0 +1,45 @@
+var React = require('react');
```
aroc: For completely new files, I think we should opt to use es2017+ syntax. This includes things like `let` and other modern language features. This is up for debate of course, but I think if we don't start moving our codebase towards that now we'll never get there without a full re-write. What's your opinion here?

client/js/app/components/explorer/query_builder/extraction_properties_filter.js
aroc: Also, we should probably upgrade this project to use `babel-latest` so we can just stay on top of the latest transpile options at all times.

client/js/app/components/explorer/query_builder/extraction_properties_filter.js
```diff
@@ -0,0 +1,45 @@
...
+      var keyStr = prevKey + '.' + key;
+      if (typeof obj[key] === 'object') {
+        return this._getKey(keyStr, obj[key]);
+      }
+
+      return keyStr;
+    }.bind(this));
+  },
+
+  render: function() {
```
aroc: Given that this component just wraps `ReactMultiSelect` and has two helper functions for formatters the props to that component, I think this component isn't required. We can format the keys as necessary inside `ExtractionOptions` and move the two functions here, `_getKeys` and `_getKey` into a utility module of some kind (or maybe simply inside `ExtractionOptions`).

client/js/app/components/explorer/visualization/keen_viz.js
```diff
@@ -35,6 +35,9 @@ var KeenViz = React.createClass({
   // ***********************
 
   shouldComponentUpdate: function(nextProps, nextState) {
+    if (this.props.model.metadata.visualization.chart_type === "table") {
```
aroc: Why do we always want to update if the viz type is a table?

client/js/app/components/common/react_multi_select.js
```diff
@@ -0,0 +1,98 @@
+var React = require('react');
```
aroc: What was the rationale for building this yourself instead of using something like `ReactSelect` (http://jedwatson.github.io/react-select/)? This component is well built btw, but being custom is means more code (including css) to maintain ourselves over time.

client/js/app/components/explorer/query_builder/extraction_properties_filter.js
joannecheng: 👍

client/js/app/components/explorer/visualization/keen_viz.js
joannecheng: `KeenViz` won't update if `extractionFields` are changed (it only updates when the viz type changes), which means when a new extraction field is selected, the table won't update. The code above will update the viz if an extraction field is selected.

aroc: Updating the entire table if not required isn't the best scenario either. You may want to check if the extraction fields have changed instead to get an accurate idea of whether we really need to update this component. Re-rendering an entire table adds to the sluggishness of the app overall.

client/js/app/components/explorer/query_builder/extraction_properties_filter.js
joannecheng: sounds good, making that change

client/js/app/components/common/react_multi_select.js
joannecheng: Since we built the other react-select components on our own, I thought that was the pattern we were following - I honestly should have asked. I was also a bit wary about using `react-select` because I wondered if it meant that we should remove our own react-select code, and replace it all with react-select (which would have meant styling, too). I'm also at the point where I feel guilty throwing more dependencies into explorer, since the project is so big already. I really wanted to use https://github.com/instructure-react/react-select-box, but it hadn't been updated in a while and some of the styling didn't work well in the context we needed, so I ended up hacking the code there and throwing it into explorer.

In retrospect, I really should have asked #web what the correct decision should have been, and I'm kind of kicking myself for that.

client/js/app/components/explorer/query_builder/extraction_properties_filter.js
joannecheng: I'm definitely up for moving to es2017 syntax. It hurt me emotionally to write `var`.

client/js/app/components/common/react_multi_select.js
aroc: I think the dependency argument here is a good one. Given that our other select components are custom, I agree the argument of following the custom pattern also has merit. I think this component is totally good to keep in here. We should probably add a few tests though.

There are probably other areas we can reduce the size of this project's build, including removing big deps like lodash in favour of modern JS native functions.

client/js/app/components/explorer/query_builder/extraction_properties_filter.js
aroc: LOL! `var` feels so bad to write nowadays eh?

PR #215, opened 2017-06-07
Fix bug with timezone offsets when saving or updating queries

aroc:
## What does this PR do?
It fixes a reported bug where queries with an absolute timeframe had the start and/or end times change when the query was saved or updated (running the query works as expected).

It also makes a core change to how Explorer handles timezones. Rather than encoding the timezone in the ISO-8601 strings, the timezone is *always* passed in the separate `timezone` parameter. This ensures the API can intelligently handle daylight savings time.


jacklittleton:
Circle CI says tests are failing, but that's not what I see when I run locally (all pass).

The time zone picker shows up as a negative number:

image 2017-06-07 at 8 11 31 am public

Then after doing a query, clicking in the address bar, and pressing Enter, the times shift and the timezone changes to UTC:

image 2017-06-07 at 8 13 15 am public

jacklittleton:
>  the timezone is always passed in the separate timezone parameter

👍 👍 👍

PR #219, opened 2017-06-14
Add automatic timezone detection (back) to Explorer

aroc:
## What does this PR do?
It adds the automatic timezone detection back to Explorer. It was removed as part of other bug fixes.


client/js/app/actions/ProjectActions.js
```diff
@@ -27,7 +27,7 @@ var ProjectActions = {
 
   fetchProjectCollections: function(client) {
     var project = ProjectStore.getProject();
-    if (!project) throw new Error("Cannot fetchProjectCollections: No project model has been created yet.");
+    if (!project) console.error("Cannot fetchProjectCollections: No project model has been created yet.");
```
dustinlarimer: Nice change :+1:

client/js/app/utils/ProjectUtils.js
```diff
@@ -139,18 +239,27 @@ module.exports = {
...
+    var isDST = DateUtils.isDST();
+    var localOffset = module.exports.getLocalTimezoneOffset();
+    var zones = CONSTANTS.TIMEZONES.filter((zone) => {
+      if (isDST) {
+        return zone.dst_offset === localOffset;
+      } else {
+        return zone.offset === localOffset;
+      }
     });
-    return found ? found.value : offset * -60;
```
dustinlarimer: The previous `offset` value here would sub in the numeric offset for the locale when a named timezone wasn't present.. would this behavior still be present in this new update?

aroc: Fixed!

PR #220, opened 2017-06-16
Fix bugs with extraction fields feature (and use make use of property_names)

aroc:
## What does this PR do?

This PR fixes a few critical bugs with the extraction fields feature:

1) It ensures the fields are cleared when the event collection is changed. This prevents the user from getting into a bad state when changing event collections and having extraction fields selected.
2) It fixes a big where the table visualization would re-render every time the user interacted with the app.
3) It fixes a UX issue where extraction fields live updated the table visualization and did not require a re-run of the query. All of our other query interactions require setting Query Builder fields and then re-running the query to see the result. The departure from this interaction style for this feature made the interaction set inconsistent. Instead of live filtering as fields are selected, this PR changes the slow to instead build up a `property_names` array that is then passed along with the query when the user clicks "Run Query". This also means we let the API handle filtering the extraction properties for us.

### Note: Please give this PR a thorough review, including extensive actual app usage to ensure it behaves as expected.


joannecheng:
👍 Thanks for fixing this bug.

aroc:
Thanks everyone! I'll get this deployed to keen.io.

PR #224, opened 2017-07-11
Use query_name as fallback name when display_name does not exist

sfalkoff:
## What does this PR do? 

This PR adds an additional fallback name for saved (and cached) queries displayed in the "Browse" tab.

When a saved/cached query does not have `metadata` or a `display_name`,  the `query_name` will be used to name the query instead. Placeholder text (i.e. "Query not named") will be displayed if none of the previously mentioned properties exist.

screen shot 2017-07-11 at 2 58 46 pm

## Why?

After creating a [saved query ](https://keen.io/docs/api/#creating-a-saved-query) or [cached query](https://keen.io/docs/api/#creating-a-cached-query) via the API, it currently appears as "Query not named" in the "Browse" tab of the query pane, making the query difficult to locate in the UI. This happens because API-generated saved/cached queries do not have a `display_name` property, or a `metadata` property (currently used for naming in the UI). 

## Testing
Tests have been added. They're passing for me locally - looking into why CI is failing after pushing the changes.


sfalkoff:
https://github.com/keen/explorer/pull/225 resolves the issue preventing CI from passing in this branch.

PR #226, opened 2017-08-02
Fixes issues that occur when saved queries are created using the API

jacklittleton:
## What it does

This builds upon the work in [PR 224](https://github.com/keen/explorer/pull/224):

`ExplorerUtils.formatQueryParams` was updated to set the `metadata` object property such that:

* `metadata` is always an object
* `metadata.display_name` is always set (not null) and defaults to the same value as `query_name` if and only if `display_name` wasn't previously set
* `visualization` is a descendant of `metadata` and is always an object
* `visualization.chart_type` is always set (not null) and defaults to 'metric' if and only if `chart_type` wasn't previously set

Added a safety valve in `BrowseQueries.buildList` to check for the existence of the `metadata` object property before referencing it.

Updated the test `should keep all explorer attributes` so that it reflects that a null `display_name` is now set to the `query_name`.

Added six new tests to check that various settings (or lack thereof) of `metadata` and descendants are properly populated by `ExplorerUtils.formatQueryParams()`

## Why it does it

When a saved query is created using the API, the resulting query structure does not have a `metadata` property. This property is an object that contains the display name and chart type of the query. The lack of this information will cause a display issue in the Browse tab of the Explorer, and console errors to appear due to the missing chart type.

_NOTE: The service that creates the query object should be modified to require at least the chart type to be a required parameter._

## How do make sure it does what it's supposed to do

Run the tests by first executing `npm run dev` in one terminal window, and then running `npm run test` in another terminal window (waiting for the initial run of `npm run dev` to complete first. Ensure no test errors are present.

To check out the new behavior for yourself, create a new saved query using the API ([see here](https://keen.io/docs/api/#creating-a-saved-query)).

Next, open a browser, open dev tools, go to your org's Explorer and then click on the "Browse" tab. Now look for the saved query you just created, and then click on it. A metric view (probably displaying "0") will appear, and there should not be any errors in the dev tools console.


adamkasprowicz:
@jacklittleton Thank you very much for your contribution!

PR #19, opened 2015-08-17
Remove master key

josephwegner:
Removes the master key requirement

@hex337 @dustinlarimer @aroc I'm gonna go ahead and merge this on the public version, for the sake of time. Any changes from private review can be moved back over.

PR #237, opened 2018-08-27
Remove the master key from the readme

### LINKED ISSUE #214: About the necessity of Master keys
Quick question (I tried to search in the docs, but didn't find it):

It's required, when we load the explorer the master key, but isn't the master key too much access to give "publicly" to users, specially when they have "scoped" access only to their organisation?

I saw the pull request #19, which I believe removed the necessity for Master key, is that correct?

Thank you

stof:
Closes #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,14 +13,13 @@ import KeenExplorer from 'keen-explorer';
 const myExplorer = new KeenExplorer('#keen-explorer-html-element')
   .client({
     projectId: 'PROJECT_ID',
-    readKey: 'READ_KEY',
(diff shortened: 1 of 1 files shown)


stof:
this also change links to use HTTPS everywhere

adamkasprowicz:
MasterKey is still used when creating Saved or cached queries in the Explorer. We're working on read-only version of the Explorer. When it will be merged in the next few weeks - I will merge this PR too. Thanks!

stof:
OK, so #19 removed the need for the master key, but a need for it was added again later when adding support for saved queries. Good to know.

PR #371, opened 2020-07-06
Refactor/filters

apryka:


lib/js/app/queryCreator/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+    return FILTER_OPERATORS.filter(operator => operator.dataTypes.includes(type));
+  }
+
+  const [state, filtersDispatcher] = useReducer(filtersReducer, []);
+  const stateRef = useRef(state);
+
+  const filtersRef = useRef(filters);
+
+  useEffect(() => {
```
maciejrybaniec, line 65: I'm not sure if we need to perform `shallowEqual` for filters. As they cannot be changed in time by selecting another query. 

So during the component mount we could just set local reducer state.

lib/js/app/queryCreator/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+      stateRef.current = state; console.log('use state');
+    }
+  }, [state])
+
+  const getInputValue = (filter: Filter) => {
+    if (filter?.propertyValue && typeof filter?.propertyValue === 'string') return filter.propertyValue;
+    return '';
+  };
+
+  const getSelectValue = (options: any, filter: Filter) => {
```
maciejrybaniec, line 87: Extract it to <FilterValue> component and add unit tests for specific UI component related with operator and property type.

lib/js/app/queryCreator/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,266 @@
...
+      <Input
+        variant="solid"
+        placeholder="Value"
+        value={getInputValue(filter)}
+        onChange={(e) => filtersDispatcher(updateFilter(idx, { propertyValue: e.target.value }))}
+      />
+    );
+  }
+
+  const getPropertyType = (item: Filter) => {
```
maciejrybaniec, line 171: Extract it from `render` function to avoid re-definition on each re-render.

lib/js/app/queryCreator/components/Filters/utils/getTypeFromValue.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (value === null) return 'null';
+  return value.constructor.name.toLowerCase();
+}
+
+const isList = (value) => {
+  const strValue = value.toString();
+  return strValue.split(',').length > 1
+}
+
+export const getTypeFromValue = (filter: Filter) => {
```
maciejrybaniec, line 14: Please, add test cases for detection of proper type based on filter value.

lib/js/app/queryCreator/config.ts
```diff
@@ -106,4 +106,5 @@ export const FIELDS_CONFIG: QueryCreatorConfig = {
   contentEncoding: ['extraction'],
   propertyNames: ['extraction'],
   latest: ['extraction'],
+  filters: ['*'],
```
maciejrybaniec, line 109: All types of analysis - instead of `funnel`.

lib/js/app/queryCreator/components/Filters/utils/convertSchemaProp.ts
```diff
@@ -0,0 +1,15 @@
+import { SchemaProp } from '../types';
+
+export const convertSchemaProp = (prop: SchemaProp) => {
+  const schemaProps = {
```
maciejrybaniec, line 4: Move away from `function` scope to avoid re-definition for each function call.

lib/js/app/queryCreator/App.tsx
```diff
@@ -115,6 +116,11 @@ const App: FC<Props> = ({ onPreviewCollection }) => {
       )}
       {showField('interval', analysis) && <Interval />}
       {showField('limit', analysis) && <Limit />}
+      {showField('filters', analysis) && (
+        <Accordion renderHeader={() => <div>Filters</div>}>
```
maciejrybaniec, line 120: Mount component with Filters editor only when user want to edit filter values.

PR #375, opened 2020-07-28
feat: 🎸 radio-group component

apryka:
https://metakeen.atlassian.net/browse/FRON-151


lib/js/app/queryCreator/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,62 @@
...
+            onChange={e => onChange(e, radio.value)}
+          />
+          <StyledLabel
+            data-testid={`${radio.value}${radio.value === value ? '-active' : ''}`}
+            htmlFor={radio.value}>
+              {radio.label}
+            </StyledLabel>
+        </Radio>
+      ))}
+      {activeRadioRef && <Marker {...markerMotion } />}
```
dariuszlacheta, line 57: `actionRadioRef` is not always true?

apryka: nope, it's true if active element is set

PR #385, opened 2020-08-11
feat: 🎸 Percentile

dariuszlacheta:


lib/js/app/queryCreator/components/Percentile/Percentile.tsx
```diff
@@ -32,16 +56,32 @@ const Percentile: FC<Props> = ({ value, onReset, onChange }) => {
...
+      {
+        tooltip && (
+          <TooltipContainer
+          {...tooltipMotion}
+            initial={{ opacity: 0}}
+            animate={{
+              opacity: 1
+            }}>
+            <Tooltip mode="dark" hasArrow={false}>
+              Percentile value has to be from range 0-100
```
maciejrybaniec, line 70: Please move text copies to separate `text.json` file to keep consistency in codebase.

lib/js/app/queryCreator/components/QueryArguments/QueryArguments.tsx
```diff
@@ -34,10 +37,11 @@ const App: FC<Props> = () => {
   const timeframe = useSelector(getTimeframe);
   const targetProperty = useSelector(getTargetProperty);
   const timezone = useSelector(getTimezone);
+  const percentile = useSelector(getPercentile);
 
   return (
     <Container>
-      <MenuItem>
+      <MenuItem width="25%">
```
maciejrybaniec, line 44: Please, correct me if I am wrong but Is defining a percentage width for each `<MenuItem>` component is necessary ? From my point of view flex should adjust all items automatically and the Percentile component could have static width.

PR #390, opened 2020-08-13
Feat/vizualize query result

apryka:
https://metakeen.atlassian.net/browse/FEE-3


lib/js/app/components/VisualizationPlaceholder/VisualizationPlaceholder.test.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import VisualizationPlaceholder from './VisualizationPlaceholder';
+
+import text from './text.json';
+
+test("render loading placeholder when loading is true", () => {
```
maciejrybaniec, line 8: `renders placeholder with loading indicator`

lib/js/app/components/VisualizationPlaceholder/VisualizationPlaceholder.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const { queryByText } = render(<VisualizationPlaceholder isLoading />);
+  const loadingPlaceholder = queryByText(text.loadingPlaceholder);
+  const placeholder = queryByText(text.placeholder);
+  expect(loadingPlaceholder).toBeInTheDocument();
+  expect(placeholder).toBeNull();
+});
+
+test("render placeholder when loading is done", () => {
```
maciejrybaniec, line 16: `renders visualization placeholder`

lib/js/app/modules/queries/reducer.ts
```diff
@@ -20,6 +21,7 @@ export const initialState: ReducerState = {
   isSavingQuery: false,
   saved: [],
   isLimited: false,
+  limitReached: false,
```
maciejrybaniec, line 24: Right now there is `isLimited ` -> which indicates the cache possibilities and `limitReached `. For external developer this names are not self-descriptive. 

Please rename `limitReached` -> `queriesExecutionLimitReached `. For caching user story we will also rename `isLimited` to something more meaningful.

lib/js/app/modules/queries/types.ts
```diff
@@ -143,4 +152,5 @@ export type QueriesActions =
   | RunQuerySuccessAction
   | GetSavedQueriesAction
   | GetSavedQueriesSuccessAction
-  | GetSavedQueriesErrorAction;
+  | GetSavedQueriesErrorAction
+  | SetQueryLimitReachedAction;
```
maciejrybaniec, line 156: Is that action invoked anywhere ? I believe to cover use case with execution queries limit reached. You should dispatch that action in error handler for run query saga.

maciejrybaniec:
```
function* runQuery(action: RunQueryAction) {
  try {
    const {
      payload: { body },
    } = action;
    const client = yield getContext('keenClient');
    const responseBody = yield client.query(body);

    yield put(runQuerySuccess(responseBody));
  } catch (error) {
    console.log('error', error);
    yield put(runQueryError(error));
  }
}

```

PR #402, opened 2020-08-26
refactor: 💡 OrderBy component

apryka:
https://metakeen.atlassian.net/browse/FEE-6


lib/js/app/queryCreator/components/OrderBy/components/OrderByProperty/OrderByProperty.component.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+import { Direction } from '../../types';
+
+type Props = {
+  /** Properties tree */
+  properties: Record<string, string[] | Record<string, any>>;
+  /** Property */
+  property?: string;
+  /** Direction */
+  direction?: Direction;
```
maciejrybaniec, line 38: `direction` is not self-descriptive. Could You please rename that property to something more meaningful like for eg. orderDirection ?

maciejrybaniec (changes requested):

PR #405, opened 2020-08-28
Refactor Schema Serialize

maciejrybaniec:
- Compute schema tree async to do not block user interface rendering.
- Remove obsolete request for schema before filters transformation.

### DIFF
--- a/lib/js/app/queryCreator/modules/events/actions.ts
+++ b/lib/js/app/queryCreator/modules/events/actions.ts
@@ -14,7 +14,7 @@ export const setEventsCollections = (collections: string[]): EventsActions => ({
   payload: { collections },
 });
 
-export const schemaComputed = (
+export const computeSchemaSuccess = (
   collection: string,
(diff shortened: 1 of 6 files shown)


lib/js/app/queryCreator/modules/events/actions.ts
```diff
@@ -4,15 +4,27 @@ import {
...
 
-import { EventsActions } from './types';
+import { EventsActions, CollectionSchema } from './types';
 
 export const setEventsCollections = (collections: string[]): EventsActions => ({
   type: SET_EVENTS_COLLECTIONS,
   payload: { collections },
 });
 
+export const schemaComputed = (
```
apryka, line 17: what do you think about putting the verb first if it's an action - `computeSchema`

maciejrybaniec: Changed to _computeSchemaSuccess_ to keep common convention across codebase.

apryka (approved):

PR #418, opened 2020-09-10
feat: 🎸 share query url feature

maciejrybaniec:


lib/js/app/modules/app/actions.ts
```diff
@@ -73,6 +75,16 @@ export const queryEditorMounted = (): AppActions => ({
...
 
+export const explorerMounted = (): AppActions => ({
+  type: EXPLORER_MOUNTED,
+});
+
+export const notificationsMounted = (): AppActions => ({
+  type: NOTIFICATIONS_MOUNTED,
+});
+
+NOTIFICATIONS_MOUNTED;
```
dariuszlacheta, line 86: IMO this is not needed here

apryka (approved):

PR #419, opened 2020-09-10
chore: 🤖 handle errors on query delete

apryka:


lib/js/app/modules/queries/saga.ts
```diff
@@ -174,16 +177,35 @@ function* deleteQuery(action: DeleteQueryAction) {
...
+      yield put(selectFirstSavedQuery());
+
+      yield notificationManager.showNotification({
+        type: 'error',
+        message: text.queryDeleted,
+        autoDismiss: true,
+      });
+    }
+
+    if (status === 500) {
```
maciejrybaniec, line 200: The code `500` could be imported from 'http-status-codes` package.

```
import HttpStatus from 'http-status-codes';

if (status === HttpStatus.INTERNAL_SERVER_ERROR) { }
```

maciejrybaniec (approved):

maciejrybaniec (approved):

PR #429, opened 2020-09-24
feat: 🎸 export results to image/JSON

apryka:
* https://metakeen.atlassian.net/browse/FEE-71
* https://metakeen.atlassian.net/browse/FEE-72


lib/js/app/modules/app/saga.ts
```diff
@@ -246,6 +249,32 @@ export function* appStart({ payload }: AppStartAction) {
...
+    fileName = `${query.analysis_type}-${query.event_collection}-${Date.now()}`;
+  }
+  return fileName;
+}
+
+export function* exportChartToImage() {
+  const node = document.getElementById('query-visualization');
+  if (!node) throw new Error('Query visualization container is not available');
+  const fileName = yield generateFileName();
+  exportToImage({ fileName, node });
```
maciejrybaniec, line 269: As we here depend on external library I suggest to wrap `exportToImage` with try catch block and handle exception with toast notification that informs user about failure export operation.

lib/js/app/utils/exportToJson.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  if (!hasExtension) {
+    downloadFileName = `${fileName}.json`;
+  }
+
+  try {
+    download(JSON.stringify(data), downloadFileName, 'text/json;charset:utf-8');
+  } catch (err) {
+    alert('There was a problem with JSON export');
+  }
```
maciejrybaniec, line 23: Please, remove try catch block and move this exception handler to `exportChartToJson` saga level. This will allow to show a toast notification about failure operation.

lib/js/app/modules/app/saga.ts
```diff
@@ -246,6 +249,32 @@ export function* appStart({ payload }: AppStartAction) {
...
+}
```
maciejrybaniec, line 270: Please talk with Gosia about communication and message in context of failure export operations.

maciejrybaniec (changes requested):

maciejrybaniec (approved):

PR #458, opened 2020-10-16
Feat/api resource url

apryka:
https://metakeen.atlassian.net/browse/FEE-80


lib/js/app/components/ActionsMenu/ActionsMenu.tsx
```diff
@@ -51,15 +56,15 @@ const ActionsMenu: FC<Props> = ({ isNewQuery, onRemoveQuery, onHideMenu }) => {
           onMouseEnter={() => !queryResults && showTooltip(true)}
           onMouseLeave={() => tooltip && showTooltip(false)}
         >
-          <AnimatePresence>
-            {tooltip && (
+          {tooltip && (
```
maciejrybaniec, line 59: Based on `framer-motion` docs the _<AnimatePresence />_ component should be mounted and conditionally render it children. By this change we'll always create a new instance of Framer <AnimatePresence /> component if tooltip is visible.

lib/js/app/modules/app/saga.ts
```diff
@@ -417,6 +420,29 @@ export function* downloadCodeSnippet({ payload }: DownloadCodeSnippetAction) {
   }
 }
 
+export function* copyApiResourceUrl({ payload }: CopyApiResourceUrlAction) {
+  const { config } = payload;
+  const query = yield select(getQuerySettings);
+  const notificationManager = yield getContext(NOTIFICATION_MANAGER_CONTEXT);
+  try {
+    const url = createResourceUrl({ query, config });
+    console.log(url);
```
maciejrybaniec, line 429: Obsolete console.log

maciejrybaniec (approved):

PR #497, opened 2020-11-06
chore: 🤖 add missing tooltips

apryka:
https://metakeen.atlassian.net/browse/FEE-195

### DIFF
--- a/lib/js/app/queryCreator/components/TargetProperty/TargetProperty.styles.ts
+++ b/lib/js/app/queryCreator/components/TargetProperty/TargetProperty.styles.ts
@@ -1,6 +1,5 @@
 import styled from 'styled-components';
 import { motion } from 'framer-motion';
-import { colors } from '@keen.io/colors';
 
 export const Container = styled.div`
   position: relative;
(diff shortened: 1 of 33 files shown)


lib/js/app/queryCreator/components/TargetProperty/TargetProperty.styles.ts
```diff
@@ -7,3 +9,17 @@ export const Container = styled.div`
 export const PropertyOverflow = styled.div`
   overflow: hidden;
 `;
+
+export const TooltipContent = styled.div`
```
maciejrybaniec, line 13: There is ~24 references to _TooltipContent_ in this code. I think this is right now a candidate to a separate reusable component. Could You please refactor it to create a reusable single component with tooltip and it's content ?

maciejrybaniec, line 13: There is ~24 references to TooltipContent in this code. I think this is right now a candidate to a separate reusable component. Could You please refactor it to create a reusable single component with tooltip and it's content ?

maciejrybaniec (changes requested):

maciejrybaniec (approved):

PR #504, opened 2020-11-17
refactor: 💡 extract <QueryCreator/>

apryka:


package.json
```diff
@@ -49,7 +49,7 @@
     "@keen.io/icons": "^1.2.1",
     "@keen.io/parser": "^2.1.1",
     "@keen.io/pubsub": "^1.1.0",
-    "@keen.io/query-creator": "^1.0.7",
+    "@keen.io/query-creator": "^1.0.8",
```
maciejrybaniec, line 52: We need to get rid of not required dependencies - that are right now part of query creator package.

maciejrybaniec (changes requested):

maciejrybaniec (approved):

PR #583, opened 2021-03-25
fix: 🐛 Widget picker z-index changed to be bigger than range slider z-index

mczapkowicz:
Closes: [FEE-505](https://metakeen.atlassian.net/browse/FEE-505)


mczapkowicz:
Fixed in: https://github.com/keen/keen/pull/850

PR #607, opened 2021-04-21
chore: 🤖 increase css specificity

apryka:


package.json
```diff
@@ -88,7 +91,11 @@
     "styled-system": "5.1.5",
     "uuid": "^8.2.0"
   },
+  "resolutions": {
+    "@types/react": "^16.9.35"
```
maciejrybaniec, line 95: The `@types/react` should be obsolete here as react package have it own types definitions.

package.json
```diff
@@ -179,5 +187,6 @@
   ],
   "publishConfig": {
     "access": "public"
-  }
+  },
+  "types": "./typings/index.d.ts"
```
maciejrybaniec, line 191: This small change will break type interference as during build in CI only dist directory is persisted to the workspace.

maciejrybaniec (changes requested):

maciejrybaniec:
With merge of this - let's wait till BI release - as it introduces a massive amount of changes in build.

PR #616, opened 2021-04-28
chore: 🤖 Share query migrated from dropdown to separate button

mczapkowicz:
[FEE-481](https://metakeen.atlassian.net/jira/software/projects/FEE/boards/10?selectedIssue=FEE-481)


src/components/BrowserQueryMenu/BrowserQueryMenu.tsx
```diff
@@ -41,6 +45,8 @@ const BrowserQueryMenu: FC<Props> = ({ onEditQuery, onRemoveQuery }) => {
   const [actionsMenu, setActionsMenuVisibility] = useState(false);
   const [actionsTooltip, showActionsTooltip] = useState(false);
   const [settingsTooltip, showSettingsTooltip] = useState(false);
+  const [shareTooltip, showShareTooltip] = useState(false);
```
apryka, line 48: what do you think about refactoring these `useStates` into one?

apryka: e.g `const [visibile, setVisible] = useState<'menu' | 'actions' | 'settings' | 'share'>(false)`

mczapkowicz: Thanks, good idea to refactor this. I changed menu tooltips to use MousePositionedTooltips component, so now it looks  much simpler.

src/components/EditorNavigation/EditorNavigation.styles.ts
```diff
@@ -54,19 +53,11 @@ export const Menu = styled.div`
   display: flex;
   margin-left: auto;
 
-  ${MenuItem} + ${MenuItem} {
+  div + div {
```
apryka, line 56: might cause problems as we have almost every component rendered as div, so IMO we should be more specific

```
export const Menu = styled.div`
  display: flex;
  margin-left: auto;

  > div + div {
    margin-left: 10px;
  }
`;
```

or

```
export const Menu = styled.div`
  display: flex;
  margin-left: auto;
  column-gap: 10px;
`;
```

mczapkowicz: Good idea, I have changed it.

apryka (approved):

PR #620, opened 2021-04-29
test: 💍 e2e tests for extraction modal

apryka:


mczapkowicz (approved):

cypress/integration/editor/Extraction.feature
```diff
@@ -0,0 +1,27 @@
...
+    And Select Event stream
+
+  @smoke-test
+  Scenario: Perform test for extracting a full collection of properties
+    When I click on "Preview events" button
+    Then Modal for "Large amount of properties" appears
+    And Close the modal
+
+  @smoke-test
+  Scenario: Perform test for selecting more properties than confirmExtractionLimit
```
maciejrybaniec, line 17: The scenario description should be written in `business` way without any implementation details like for eg. reference to variable names etc.
So this example scenario description could be like this: Shows extraction confirmation modal when count of properties exceed the limit

maciejrybaniec (changes requested):

maciejrybaniec (approved):

PR #692, opened 2021-09-08
feat: 🎸 Table chart formatting

mczapkowicz:
Part of the solution for: https://metakeen.atlassian.net/jira/software/projects/FEE/boards/10?selectedIssue=FEE-599


src/components/Editor/Editor.tsx
```diff
@@ -109,6 +111,15 @@ const Editor: FC<Props> = ({
     )
   );
 
+  useEffect(() => {
```
maciejrybaniec, line 114: Could You elaborate why this `useEffect` is required here ? The widget customization input settings are serialized during first render.

maciejrybaniec (changes requested):

maciejrybaniec:
Looks good for me. Before we merge it - we need to wait until back-end team add support for properties with `dot` notation.

maciejrybaniec (approved):

PR #694, opened 2021-09-21
feat: 🎸 Routing added

mczapkowicz:


src/KeenExplorer.tsx
```diff
@@ -49,13 +57,28 @@ export class KeenExplorer {
...
     createI18n(translationSettings);
 
+    const initialView = props.initialView || 'browser';
+
+    let initialViewUrl = INITIAL_VIEWS[initialView];
+    if (savedQuery) {
+      initialViewUrl += '?savedQuery=' + savedQuery;
+    }
+
+    console.log('initialViewUrl', initialViewUrl);
```
apryka, line 76: I think this is obsolete

src/constants.ts
```diff
@@ -57,3 +57,13 @@ export const TOOLTIP_MOTION = {
...
 
 export const DEFAULT_TIMEZONE_FOR_QUERY = 'Etc/UTC';
+
+export const ROUTES: Record<string, string> = {
+  BROWSER: '/browser',
+  EDITOR: '/editor',
+};
+
+export const INITIAL_VIEWS: Record<string, string> = {
+  browser: '/browser',
```
apryka, line 67: `ROUTES.BROWSER` maybe? so the future changes may be applied into one place

src/middleware/viewUpdate.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Creates route update middleware that notifies container applications
+ * @param updateHandler - Update view handler
+ *
+ * @return Middleware - redux middleware
+ *
+ */
+
+const ROUTES_MAP: Record<string, Views> = {
+  '/browser': 'browser',
```
apryka, line 14: there are only 2 routes here but I think it's better to use `ROUTES` here

apryka (dismissed):

src/components/App.tsx
```diff
@@ -131,31 +135,46 @@ class App extends Component<Props> {
...
+                onRunQuery={() =>
+                  this.props.runQuery({ query: this.props.query })
+                }
+                onSelectQuery={(queryName) => {
+                  this.props.resetQueryResults();
+                  this.props.selectSavedQuery(
+                    queryName,
+                    this.props.autorunQuery
+                  );
+                  this.props.push('/browser?savedQuery=' + queryName);
```
maciejrybaniec, line 153: Instead of inline route definition with `browser` - could we use here `ROUTES` constants and inline savedQuery param.

src/components/BrowserNavigation/BrowserNavigation.test.tsx
```diff
@@ -48,6 +48,10 @@ test('allows user to create a new query', () => {
 
   expect(store.getActions()).toMatchInlineSnapshot(`
     Array [
+      Object {
```
maciejrybaniec, line 51: I checked `createNewQuery` saga and `resetSavedQuery` is part of saga flow.
Do we need to additionally perform `dispatch(createNewQuery());` in BrowserNavigation component ?

```
export function* createNewQuery() {
  yield put(setViewMode('editor'));
  const pubsub = yield getContext(PUBSUB_CONTEXT);
  yield pubsub.publish(NEW_QUERY_EVENT);

  yield put(queriesActions.resetQueryResults());
  yield put(resetVisualization());
  yield put(savedQueryActions.resetSavedQuery());
}
```

maciejrybaniec (changes requested):

apryka (approved):

src/modules/savedQuery/saga/fetchSavedQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+import { SavedQueryAPIResponse } from '../../../types';
+import {
+  KEEN_CLIENT_CONTEXT,
+  NOTIFICATION_MANAGER_CONTEXT,
+} from '../../../constants';
+import { fetchSavedQuery as fetchSavedQueryAction } from '../actions';
+import { savedQueryActions } from '../index';
+import { queriesActions } from '../../queries';
+
+export function* fetchSavedQuery({
```
maciejrybaniec, line 11: As we discussed - the `fetchSavedQuery` saga should be responsible only for fetching saved query definition.
The missing part is rehydrating the store with saved query name, tags and visualization settings.

maciejrybaniec (changes requested):

src/components/App.tsx
```diff
@@ -131,31 +135,46 @@ class App extends Component<Props> {
...
-              this.props.selectSavedQuery(queryName, this.props.autorunQuery);
-            }}
-            onEditQuery={(queryName) => {
-              this.props.editQuery(queryName);
-            }}
+        <Switch>
+          <Route
+            exact
+            path={ROUTES.BROWSER}
+            render={() => (
```
maciejrybaniec, line 142: Just a suggestion - but in my opinion using `children` instead of `render` could make this code a bit cleaner by using composition.

maciejrybaniec (approved):

src/components/App.tsx
mczapkowicz: Good idea

